PR #4, opened 2024-04-04
Add API and CLI command for updating multiple indexes at one

ludwiktrammer:


docs/how-to/update_similarity_indexes.md
```diff
@@ -0,0 +1,88 @@
...
+from db_ally import SimilarityIndex
+
+# Create a similarity index
+similarity_index = SimilarityIndex(fetcher=fetcher, store=store)
+
+# Update the similarity index
+await similarity_index.update()
+```
+
+### Detect Similarity Indexes in Views
```
mhordynski, line 58: Maybe it would be beneficial to add extra method on collection to update all indexes used inside? Something like Collection.update_indexes() which aggregates all SimilarityIndexes and executes update on them

ludwiktrammer: Maybe. I viewed the CLI as the primary method of updating collections, and focused less on ways of doing this in code that can't be used by CLI, but you may be right. On the other hand this seems like a separate area, so how about we open a new ticket and do this in a new PR?

mhordynski (approved):

PR #7, opened 2024-04-09
API Reference continuation

ds-sebastianchwilczynski:


docs/reference/event_handlers/index.md
```diff
@@ -1,20 +1,47 @@
...
+    To learn how to create a cutom `EventHandler`, visit [How-To: Create your own event handler](../../how-to/create_custom_event_handler.md).
+
+
+
 
 ## Lifecycle
 
 
-Every run of db-ally will trigger any instance of EventHandler that was registered using `dbally.use_event_handler` method in following manner:
+Every run of [dbally.Collection.ask](../collection.md/#collection.ask) will trigger every instance of EventHandler that was registered using [`dbally.use_event_handler`](../index.md/#dbally.use_event_handler) method in following manner:
```
ludwiktrammer, line 14: The sentence structure seems a bit complicated. Maybe "Each run of ... will trigger all instances of ..."?

docs/reference/event_handlers/index.md
```diff
@@ -1,20 +1,47 @@
...
+
 
-1. `EventHandler.request_start` is called with [RequestStart](#dbally.data_models.audit.RequestStart), it can return a context object that will be passed to next calls.
-2. For each event that occurs during the run, `EventHandler.event_start` is called with an Event and the context object returned by `EventHandler.request_start`. It can return context for the event_end method.
-3. When the event ends `EventHandler.event_end` is called with an Event and the context object returned by `EventHandler.event_start`.
-4. On the end of the run `EventHandler.request_end` is called with [RequestEnd](#dbally.data_models.audit.RequestEnd) and the context object returned by `EventHandler.request_start`.
+``` mermaid
```
ludwiktrammer, line 17: I love that you added a diagram but I would prefer it to be alongside the description not instead of it.

docs/reference/iql/index.md
```diff
@@ -2,12 +2,32 @@
...
+        - name
+        - arguments
 
 ::: dbally.iql.syntax.BoolOp
+    options:
+        members:
+        - match
 
 ::: dbally.iql.syntax.And
+    options:
```
ludwiktrammer, line 21: Is the "members" option defined everywhere because you want to avoid subclasses having members from their parent classes listed?

I'm not sure about this choice - it's nice to see everything that can be used when using a particular class, no matter where it was originally defined.

And even if you want to hide members from parent classes - we should look for an automatic option. Doing this by hand using a whitelist seems extremely error prone - it almost guarantees that the documentation will become obsolete as new members are added to different classes.

src/dbally/audit/event_handlers/base.py
```diff
@@ -10,37 +10,44 @@
...
         """
-        Log the start of the request.
+        Function that is called at the beginning of every `Collection.ask` execution.
 
         Args:
-            user_request: The start of the request.
+            user_request: Object containing name of collection and asked query
+
+        Returns:
+            Context that can be of `Any` type, which is passed to the future callbacks
```
ludwiktrammer, line 25: `Any` is part of static type analysis in Python, which our users doesn't necessarily have to use and understand.

I would explain this in words. And probably I wouldn't even say that the method can return "any type" because the type has to be consistent for a given implementation (that's why it is defined here as `RequestCtx`, not `Any`). I would probably describe it as an "implementation-specific request context object".

src/dbally/audit/event_handlers/base.py
```diff
@@ -10,37 +10,44 @@
...
+            Context that can be of `Any` type, which is passed to the `event_end` callback
         """
 
     @abc.abstractmethod
     async def event_end(
         self, event: Union[None, LLMEvent], request_context: RequestCtx, event_context: EventCtx
     ) -> None:
         """
-        Log the end of the event.
+        Function that is called during every event execution. For now we treat every LLM call as an event.
```
ludwiktrammer, line 47: "For now we treat every LLM call as an event." - this doesn't seem to describe the class itself, just a detail of how it is currently use, which is not an inherent property of the class and will probably change soon. I wouldn't include it here.

src/dbally/audit/event_handlers/cli_event_handler.py
```diff
@@ -17,7 +17,21 @@
...
+    ### Usage
+
+    ```python
+        import dbally
+        from dbally.audit.event_handlers.cli_event_handler import CLIEventHandler
+
+        dbally.use_event_handler(CLIEventHandler())
+    ```
+
+    from now, during every `Collection.ask` execution you will see output similar to the one below:
```
ludwiktrammer, line 32: "from now" -> "for now" -> "currently"

src/dbally/views/pandas_base.py
```diff
@@ -33,20 +32,24 @@ async def apply_filters(self, filters: IQLQuery) -> None:
...
+        Args:
+            filters: IQLQuery object representing the filters to apply
         """
         self._filter_mask = await self.build_filter_node(filters.root)
 
     async def build_filter_node(self, node: syntax.Node) -> pd.Series:
         """
-        Converts a filter node from the IQLQuery to a Pandas Series representing
-        boolean mask to be applied to the dataframe.
+        Converts a filter node from the IQLQuery to a Pandas Series being
```
ludwiktrammer, line 42: Not that important but I preferred "representing" - the object is not inherently a mask, it is just a boolen DataFrame, that _represents_ a mask.

src/dbally/views/pandas_base.py
```diff
@@ -63,11 +66,16 @@ async def build_filter_node(self, node: syntax.Node) -> pd.Series:
...
+        Be aware that before running this method you need to execute `apply_filters`
 
-        :param dry_run: If True, the method will only return the mask that would be applied to the dataframe
+        Args:
+            dry_run: If True, the method will only add `context` field to the `ExecutionResult` with the\
+            mask that would be applied to the dataframe
 
-        :return: ExecutionResult object with the results and context information
+        Returns:
+            ExecutionResult object with the results, being a result of DataFrame.to_dict(orient="record")\
```
ludwiktrammer, line 77: The fact that we've used `DataFrame.to_dict(orient="record")` is definitely an implementation detail. It's in the same format as `ExecutionResult` object from all other view classes. How we obtained it is not important.

src/dbally/views/sqlalchemy_base.py
```diff
@@ -21,15 +21,19 @@ def __init__(self, sqlalchemy_engine: sqlalchemy.engine.Engine) -> None:
...
+        Creates the initial
+        [SqlAlchemy select object
+        ](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select)
+        which will be used to build the query.
         """
 
     async def apply_filters(self, filters: IQLQuery) -> None:
         """
-        Applies the chosen filters to the view.
+        Applies the chosen filters to the view, using the SQL `where` statement.
```
ludwiktrammer, line 33: " using the SQL `where` statement" - seems like an implementation detail

src/dbally/views/sqlalchemy_base.py
```diff
@@ -63,16 +67,16 @@ async def _build_filter_bool_op(self, bool_op: syntax.BoolOp) -> sqlalchemy.Colu
 
     def execute(self, dry_run: bool = False) -> ExecutionResult:
         """
-        Executes the generated SQL query and returns the results.
-
-        :param dry_run: If True, only generate the query without executing it
+        Executes the generated SQL query and returns the results. Be aware that before running\
```
ludwiktrammer, line 70: You don't need to. If you run `apply_filters` that the query will execute with those filters applied, if you run it without `apply_filters` that the query will also successfully execute, without applying any filters. Seems to me like common sense.

src/dbally/views/sqlalchemy_base.py
```diff
@@ -63,16 +67,16 @@ async def _build_filter_bool_op(self, bool_op: syntax.BoolOp) -> sqlalchemy.Colu
...
+        this method you need to execute `apply_filters`
 
-        :return: Query results
+        Args:
+            dry_run: If True, only fills the context field of the `ExecutionResult` with the query without executing it
```
ludwiktrammer, line 74: "only adds the SQL query to the context field without executing the query" (it doesn't really "fills the context field with the query")

docs/reference/iql/index.md
ds-sebastianchwilczynski: It's two-fold:
1. For some classes I want to hide public methods that are used by different classes, but not by the users.
2. The second use case is exactly what you wrote. Personally I prefer parent functions to be hidden if they do not implement different logic and to be shown when they differ in the implementation. But probably this is to be discussed.

I agree that this should be automated will think on how to do this.

ds-sebastianchwilczynski: Yes, so there is an [inherited_members field](https://mkdocstrings.github.io/python/usage/configuration/members/#inherited_members)

Now we have to decide whether we show them or not

src/dbally/audit/event_handlers/cli_event_handler.py
ds-sebastianchwilczynski: I ment that after using CLIEventHandler by writing "from now". But I understand the amiguity

src/dbally/views/sqlalchemy_base.py
ds-sebastianchwilczynski: Yes you are right, I somehow wanted to give a little bit more details to the users. But why should they care about it.

docs/reference/event_handlers/index.md
```diff
@@ -1,20 +1,49 @@
...
-2. For each event that occurs during the run, `EventHandler.event_start` is called with an Event and the context object returned by `EventHandler.request_start`. It can return context for the event_end method.
-3. When the event ends `EventHandler.event_end` is called with an Event and the context object returned by `EventHandler.event_start`.
-4. On the end of the run `EventHandler.request_end` is called with [RequestEnd](#dbally.data_models.audit.RequestEnd) and the context object returned by `EventHandler.request_start`.
-
-::: dbally.audit.event_handlers.EventHandler
+2. For each event that occurs during the run, `EventHandler.event_start` is called with the context object returned by `EventHandler.request_start` and an Event object. It can return context for the `EventHandler.event_end` method.
+3. When the event ends `EventHandler.event_end` is called with the context object returned by `EventHandler.event_start` and an Event object.
+4. On the end of the run `EventHandler.request_end` is called with the context object returned by `EventHandler.request_start` and the [RequestEnd](#dbally.data_models.audit.RequestEnd).
```
ludwiktrammer, line 16: Here as well - the link will remain correct even if the documentation of the class gets moved to a different file

docs/reference/views/databases.md
```diff
@@ -1,9 +1,13 @@
 # Databases Base Views
 
+!!! tip
+    To learn how to use db-ally with databases see [How-To: Use SQL databases with db-ally](../../how-to/sql_views.md).
+
+
 ::: dbally.SqlAlchemyBaseView
     options:
         members:
```
ludwiktrammer, line 9: Why?

src/dbally/views/pandas_base.py
```diff
@@ -62,11 +65,15 @@ async def build_filter_node(self, node: syntax.Node) -> pd.Series:
 
     def execute(self, dry_run: bool = False) -> ViewExecutionResult:
         """
-        Executes the view and returns the results. The results are filtered based on the applied filters.
+        Executes the view and returns the results. The results are filtered based on the applied filters.\
+        Be aware that before running this method you need to execute `apply_filters`
```
ludwiktrammer, line 69: Similar comment to the one that I had on SQLAlchemy one before

docs/reference/event_handlers/index.md
```diff
@@ -1,20 +1,49 @@
...
+# EventHandler Base Class
 
 db-ally provides an `EventHandler` abstract class that can be used to log the runs of db-ally to any external systems.
 
+!!! tip
+    To learn how to create a cutom `EventHandler`, visit [How-To: Create your own event handler](../../how-to/create_custom_event_handler.md).
 
 ## Lifecycle
 
+Each run of [dbally.Collection.ask](../collection.md/#collection.ask) will trigger all instances of EventHandler registered using [`dbally.use_event_handler`](../index.md/#dbally.use_event_handler).
```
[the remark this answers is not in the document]
ds-sebastianchwilczynski: Unfortunately this seems not to work.

When I am at `/reference/event_handlers/`
and click the link it takes me to the `reference/event_handlers/dbally.Collection.ask` and return 404

So I think it is necessary :( What I can do is to remove `.md` extensions

docs/reference/iql/iql_generator.md
```diff
@@ -1,3 +1,11 @@
 # IQLGenerator
 
 ::: dbally.iql_generator.iql_generator.IQLGenerator
+    options:
+        members:
+        - generate_iql
```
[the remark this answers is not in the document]
ds-sebastianchwilczynski: Right, I'll bring all back

docs/reference/event_handlers/index.md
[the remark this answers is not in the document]
ds-sebastianchwilczynski: nvm. I am blind and didn't use double `[]`

docs/reference/collection.md
```diff
@@ -10,4 +10,4 @@
         - add
```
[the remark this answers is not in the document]
ds-sebastianchwilczynski: Sure

src/dbally/llm_client/base.py
```diff
@@ -80,9 +80,10 @@ async def call(
         Calls LLM API endpoint.
 
         Args:
-            prompt: Text to be asked.
-            response_format: Optional argument used in the OpenAI API - used to force json output
+            prompt: prompt passed to the LLM.
+            response_format: Optional argument used in the OpenAI API - used to force a json output
             options: Additional settings used by LLM.
+            event: an LLMEvent instance which fields should be filled during the method executionOptio.
```
ludwiktrammer, line 86: I believe the "Optio" part is here by mistake

ludwiktrammer (approved):

src/dbally/llm_client/base.py
ds-sebastianchwilczynski: I just tried to invent a new language, unsuccessfully :(

PR #11, opened 2024-04-10
Add Collection methods to detect/update similarity indexes

ludwiktrammer:


src/dbally/_collection.py
```diff
@@ -226,3 +228,29 @@ async def ask(self, question: str, dry_run: bool = False, return_natural_respons
...
+            (represented by a tupple containing view name, method name and argument name)
+        """
+        indexes: Dict[AbstractSimilarityIndex, List[Tuple[str, str, str]]] = {}
+        for view_name in self._views:
+            view = self.get(view_name)
+            filters = view.list_filters()
+            for filter_ in filters:
+                for param in filter_.parameters:
+                    if param.similarity_index:
+                        indexes.setdefault(param.similarity_index, []).append((view_name, filter_.name, param.name))
```
mhordynski, line 247: nit: `defaultdict` from `collections` is helpful

src/dbally/_collection.py
```diff
@@ -226,3 +228,29 @@ async def ask(self, question: str, dry_run: bool = False, return_natural_respons
...
+        return indexes
+
+    async def update_similarity_indexes(self) -> None:
+        """
+        Update all similarity indexes from all views in the collection.
+        """
+        indexes = self.get_similarity_indexes()
+        update_corutines = [index.update() for index in indexes]
+        await asyncio.gather(*update_corutines)
```
mhordynski, line 256: Hmm, in the case that any update task fails, we will terminate the others, which may cause an unexpected state in the indexes. I think we should use asyncio.gather with `return_exceptions=True` and return some sort of object from `update_similarity_indexes` with an update status or exception.

src/dbally/_collection.py
ludwiktrammer: I know of defaultdict, but since here the object is part of our public interface I have a slight preference for returning the standard dictionary, to avoid any surprises from the users regarding a non-standard behavior from the returned object (and/or having to document the non-standard behavior).

src/dbally/_collection.py
ludwiktrammer: Good catch, will do!

mhordynski (approved):
lgtm

PR #15, opened 2024-04-17
feat(views): allow for freeform views

ludwiktrammer:
Ground work for supporting a new type of views: freeform. Freeform views are not bounded by IQL and can answer natural language questions directly.


src/dbally/data_models/prompts/query_explainer_prompt_template.py
```diff
@@ -15,37 +21,30 @@ def __init__(
...
-            "of a list of filters that answers users' question. "
+            "that answers users' question. "
             "You are very intelligent and obedient.\n"
             "Your task is to provide natural language description of the table used by the logical query "
             "to the database.\n"
             "Describe the table in a way that is short and informative.\n"
-            "Make your answer as short as possible and start it by infroming the user that the table is too long "
-            "to be used in LLM call due to token limit.\n"
+            "Make your answer as short as possible and start it by infroming the user that the data is too long "
+            "to be used in LLM call due to token limit and then describe the table "
```
ludwiktrammer, line 39: In my tests the new part ("and then describe the table based on the question and the query") is required for the model to produce anything beyond "The data is too long to be used in LLM call".

Now the model produces a simplistic description of the table ("The table contains 23 employees from France").

src/dbally/data_models/prompts/query_explainer_prompt_template.py
```diff
@@ -15,37 +21,30 @@ def __init__(
...
-        self.chat = _check_prompt_variables(chat, {"question", "filters"})
+        self.chat = _check_prompt_variables(chat, {"question", "query", "number_of_results"})
 
 
-default_iql_explainer_template = IQLExplainerPromptTemplate(
+default_query_explainer_template = QueryExplainerPromptTemplate(
     chat=(
         {
             "role": "system",
             "content": "You are a helpful assistant that helps describe a table generated by a query composed "
```
mhordynski, line 32: I think `composed` doesn't make sense anymore

src/dbally/collection.py
```diff
@@ -48,7 +46,7 @@ def __init__(
         self,
         name: str,
         view_selector: ViewSelector,
-        iql_generator: IQLGenerator,
```
mhordynski, line 51: Okay, so here the idea is that rather than injecting your own IQLGenerator on collection level you can override method `get_iql_generator` on your view?

src/dbally/data_models/prompts/query_explainer_prompt_template.py
ludwiktrammer: Ok, I'll try this version.

More generally I think we should make about improving this scenario from a Data Science view point. When the model is given the data the answers are really good, but the fallback answers are extremely primitive (you could probably just make them with string substitution, without using LLM at all). For now, I wanted to avoid opening the Pandora's box, so I tried to change the prompt as little as humanly possible ;)

src/dbally/collection.py
ludwiktrammer: Yes. It's a byproduct of other changes but I think it might make a lot of sense sense. This means that the IQLGenerator implementation is declared per view (or per view type), not per collection. So for example, you can have extended IQL for a particular data source (connected with a view that knows how to handle it).

src/dbally/data_models/prompts/query_explainer_prompt_template.py
ludwiktrammer: I tried your version, seems to work reasonably okay so I left it in.

src/dbally/data_models/prompts/query_explainer_prompt_template.py
ludwiktrammer: Fixed

src/dbally/collection.py
ds-sebastianchwilczynski: My little concern is that Python programmers are probably not used to getters and setters that much. What if we kept the argument but expected a callable?

src/dbally/views/structured.py
```diff
@@ -0,0 +1,100 @@
...
+from dbally.data_models.execution_result import ViewExecutionResult
+from dbally.iql import IQLError, IQLQuery
+from dbally.iql_generator.iql_generator import IQLGenerator
+from dbally.llm_client.base import LLMClient
+from dbally.views.exposed_functions import ExposedFunction
+
+from .base import BaseView
+
+
+class BaseStructuredView(BaseView):
```
ds-sebastianchwilczynski, line 14: Why didn't we use the [prefix][base class name] convention?

It's totally fine for me to keep the current name, but that was the first thing I thought about :smile:

src/dbally/nl_responder/nl_responder.py
```diff
@@ -5,32 +5,35 @@
...
+)
 from dbally.llm_client.base import LLMClient
 from dbally.nl_responder.token_counters import count_tokens_for_huggingface, count_tokens_for_openai
 
 
 class NLResponder:
     """Class used to generate natural language response from the database output."""
 
+    # Keys used to extract the query from the context (ordered by priority)
+    QUERY_KEYS = ["iql", "sql", "query"]
```
ds-sebastianchwilczynski, line 24: What about API response? Maybe we should think about something generic that can be added to the context?

For now context basically can take a form of anything. Maybe we should think about Context dataclass or something? Context content seems to be consistent w.r.t the view type.

src/dbally/views/structured.py
ludwiktrammer: I'm not familiar with the convention. I try to name each class in a way that best describes the actual class.

src/dbally/nl_responder/nl_responder.py
ludwiktrammer: What do you mean by "the API response"? The actual data? This is specifically a prompt fot the case when there is to much data to fit in the prompt, there is a separate prompt that includes the data (which is used whenever possible).

In general, I think there is a lot that can be improved when it comes to NL Responder with a lot of data but I think it would be best to create a separate ticket about researching possible solutions.

ds-sebastianchwilczynski: Agree that separate ticket should be created. Besides this, everything clear.

mhordynski (approved):
lgtm

PR #18, opened 2024-04-22
Feat: Integration with Chromadb

ds-sebastianchwilczynski:


src/dbally/similarity/chroma_store.py
```diff
@@ -0,0 +1,96 @@
...
+from dbally.similarity.store import SimilarityStore
+
+
+class ChromadbStore(SimilarityStore):
+    """Class that stores text embeddings using [Chroma](https://docs.trychroma.com/)"""
+
+    def __init__(
+        self,
+        index_name: str,
+        client: chromadb.Client,
```
ludwiktrammer, line 15: I would rename this to `chroma_client` since there is a name clash between "Chroma client" and the "Embedding client" that almost every store takes as a constructor argument.

src/dbally/similarity/chroma_store.py
```diff
@@ -0,0 +1,96 @@
...
+    ):
+        super().__init__()
+        self.index_name = index_name
+        self.client = client
+        self.embedding_calculator = embedding_calculator
+        self.max_distance = max_distance
+
+        self._metadata = {"hnsw:space": distance_method}
+
+    def _get_collection(self) -> chromadb.Collection:
```
ludwiktrammer, line 28: The name "collection" here clashes with dbally's Collection. It would definitely try to be more explicit about this being chroma's collection - both in naming things in code and in docstrings.

src/dbally/similarity/chroma_store.py
```diff
@@ -0,0 +1,96 @@
...
+
+    async def store(self, data: List[str]) -> None:
+        """
+        Fills chroma collection with embeddings of provided string. As the id uses hash value of the string.
+
+        Args:
+            data: The data to store.
+        """
+
+        # chroma requires an unique id for every document
```
ludwiktrammer, line 64: If chroma requires an **unique** id for every document than the `hash` function is not the right solution:

1. It will produce the same hash each time the items happen to have the same contents
2. It does not guarantee no collisions and its space is relatively small 

You should probably generate uuids, or if you want hash (and the first point isn't a concern) - a proper hash function like sha256.

ludwiktrammer (changes requested):

.license-whitelist.txt
```diff
@@ -22,4 +22,6 @@ Python Software Foundation License, MIT License
 Unlicense
 Proprietary License
 Historical Permission Notice and Disclaimer (HPND)
-ISC
\ No newline at end of file
+ISC
+UNKNOWN
```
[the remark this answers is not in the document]
ds-sebastianchwilczynski: You are completely right, yes one of the chroma dependencies does not specify the licence. I'll dive into this problem and look for better solutions.

src/dbally/similarity/chroma_store.py
mhordynski: > 1. It will produce the same hash each time the items happen to have the same contents

It is actually desired behaviour for us - if contents are the same we're fine with using the same document

agree with sha256

.license-whitelist.txt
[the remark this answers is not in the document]
ds-sebastianchwilczynski: added it to `.libraries-whitelist.txt`

src/dbally/similarity/chroma_store.py
ludwiktrammer: > It is actually desired behaviour for us - if contents are the same we're fine with using the same document

I'm just going by the "requires an unique id for every document" comment. If it doesn't, the comment should probably by changed instead.

docs/how-to/use_chromadb_store.md
```diff
@@ -0,0 +1,74 @@
...
+country_similarity = SimilarityIndex(store, DummyCountryFetcher())
+
+
+@decorators.view_filter()
+def from_country(self, country: Annotated[str, country_similarity]) -> sqlalchemy.ColumnElement:
+    """
+    Filters candidates from a specific country.
+    """
+    return Candidate.country == country
+```
```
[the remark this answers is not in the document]
ds-sebastianchwilczynski: I added a reference

docs/how-to/use_chromadb_store.md
```diff
@@ -0,0 +1,70 @@
+# How-To: Use Chromadb to Store Similarity Index
+
+[ChromadbStore][dbally.similarity.ChromadbStore] allows to use [Chroma vector database](https://docs.trychroma.com/api-reference#methods-on-collection) as a store inside the [SimilarityIndex][dbally.similarity.SimilarityIndex].With this feature, when someone searches for 'Show my flights to the USA' and we have 'United States' stored in our database as the country's value, the system will recognize the similarity and convert the query from 'get_flights(to="USA")' to 'get_flights(to="United States")'.
```
mhordynski, line 3:
```suggestion
[ChromadbStore][dbally.similarity.ChromadbStore] allows to use [Chroma vector database](https://docs.trychroma.com/api-reference#methods-on-collection) as a store inside the [SimilarityIndex][dbally.similarity.SimilarityIndex]. With this feature, when someone searches for 'Show my flights to the USA' and we have 'United States' stored in our database as the country's value, the system will recognize the similarity and convert the query from 'get_flights(to="USA")' to 'get_flights(to="United States")'.
```

mhordynski (approved):
lgtm

PR #20, opened 2024-04-23
feat: add text2sql freeform view

mhordynski:
It adds new freeform view: text2sql

### DIFF
--- a/src/dbally/collection.py
+++ b/src/dbally/collection.py
@@ -240,7 +240,7 @@ class Collection:
         for view_name in self._views:
             view = self.get(view_name)
 
-            if not issubclass(view.__class__, BaseStructuredView):
+            if not isinstance(view, BaseStructuredView):
                 continue
 
             filters = view.list_filters()
--- a/src/dbally/collection.py
+++ b/src/dbally/collection.py
@@ -13,6 +13,7 @@ from dbally.nl_responder.nl_responder import NLResponder
 from dbally.similarity.index import AbstractSimilarityIndex
 from dbally.utils.errors import NoViewFoundError
 from dbally.view_selection.base import ViewSelector
+from dbally.views.base import BaseView
 from dbally.views.structured import BaseStructuredView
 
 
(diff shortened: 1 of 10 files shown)


src/dbally/collection.py
```diff
@@ -238,6 +239,10 @@ def get_similarity_indexes(self) -> Dict[AbstractSimilarityIndex, List[Tuple[str
         indexes: Dict[AbstractSimilarityIndex, List[Tuple[str, str, str]]] = {}
         for view_name in self._views:
             view = self.get(view_name)
+
+            if not issubclass(view.__class__, BaseStructuredView):
```
ludwiktrammer, line 243: `if not isinstance(view, BaseStructuredView):` would seems like more direct solution.

src/dbally/views/freeform/text2sql/_autodiscovery.py
```diff
@@ -0,0 +1,328 @@
...
+            engine=self._engine,
+            whitelist=self._whitelist,
+            blacklist=self._blacklist,
+            description_extraction=self._description_extraction,
+            similarity_enabled=self._similarity_enabled,
+            llm_client=llm_client,
+        )
+
+
+def configure_text2sql_auto_discovery(engine: Engine) -> _AutoDiscoveryBuilder:
```
ludwiktrammer, line 164: I'm not a fan of returning instances of private classes from public functions - i.e., making a private type part of the public interface. Since the users will be handling the object they may want to reference it's type. So in my mind `AutoDiscoveryBuilder` should be public as well.

src/dbally/views/freeform/text2sql/_autodiscovery.py
```diff
@@ -0,0 +1,328 @@
...
+            if isinstance(self._description_extraction, _DBCommentsDescriptionExtraction):
+                description = table.comment
+            elif isinstance(self._description_extraction, _LLMSummaryDescriptionExtraction):
+                example_rows = self._get_example_rows(connection, table, self._description_extraction.example_rows_cnt)
+                description = await self._generate_llm_summary(ddl, example_rows)
+            else:
+                raise ValueError(f"Invalid description extraction strategy: {self._description_extraction}")
+
+            if self._similarity_enabled:
+                similarity = await self._suggest_similarity_indexes(connection, description or "", table)
```
ludwiktrammer, line 266: I would probably skip the call to the LLM altogether in cases when there are more than n unique values in a given column

mhordynski: Why? TRIGRAM method can be still relevant even for columns with high cardinality. But yeah, some logic may be introduce to reduce llm calls here, I think we can have a closer look on that later when we will be implementic full integration between text2sql view and similarity indexes.

src/dbally/views/freeform/text2sql/_autodiscovery.py
mhordynski: you are right, fixed

ds-sebastianchwilczynski:
All issues encountered during running it on the BIRD benchmark seems to be resolved. As the first implementation LGTM.

I'll be testing it soon on the benchmark so you can wait a little bit with merging

ds-sebastianchwilczynski (approved):
Found no issues when running it on the BIRD benchmark

PR #26, opened 2024-05-09
refactor: move prompt templates to modules which use them

karllu3:


src/dbally/prompts/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .common_validation_utils import ChatFormat, PromptTemplateError, _check_prompt_variables, _extract_variables
 from .prompt_builder import PromptBuilder
 
-__all__ = ["PromptBuilder"]
+__all__ = ["PromptBuilder", "PromptTemplateError", "_extract_variables", "_check_prompt_variables", "ChatFormat"]
```
mhordynski, line 4: we shouldn't export private function out of the package, if they're required somewhere else we should change them to public

mhordynski:
I think `refactor:` commit type would be better choice here rather than `fix:`

src/dbally/prompts/__init__.py
karllu3: fixed
_extract_variables shall not be exposed
I have changed _check_prompt_variables to check_prompt_variables

mhordynski (approved):
lgtm 👌

PR #27, opened 2024-05-09
feat(views): enable injecting similarity indexes into text2sql freeform views

mhordynski:


src/dbally/collection.py
```diff
@@ -228,26 +227,20 @@ async def ask(self, question: str, dry_run: bool = False, return_natural_respons
...
 
-    def get_similarity_indexes(self) -> Dict[AbstractSimilarityIndex, List[Tuple[str, str, str]]]:
+    def get_similarity_indexes(self) -> Dict[AbstractSimilarityIndex, List[str]]:
         """
         List all similarity indexes from all structured views in the collection.
 
         Returns:
             Dictionary with similarity indexes as keys and values containing lists of places where they are used
-            (represented by a tupple containing view name, method name and argument name)
+            (represented by a tuple containing view name, method name and argument name)
```
ludwiktrammer, line 236: This part (about it being represented by a tuple) seems to no longer be true - from what I see, you changed it to be represented by just the name itself. I'm also not sure about the change - I think it is pretty important that one is able to locate the exact location of a given index (even for the error message). There is also the question about getting `SimilarityIndexDetector` updated to work with freeform views - it would be hard for different reasons (we should probably talk about that separately) but if we were to do it (or even use a different mechanism with similar functionality), we would definitely need the location of indexes (as the detector/cli allows user to filter indexes by those locations).

src/dbally/views/freeform/text2sql/_autodiscovery.py
```diff
@@ -144,6 +144,17 @@ def generate_description_by_llm(self, example_rows_cnt: int = 5) -> Self:
         self._description_extraction = _LLMSummaryDescriptionExtraction(example_rows_cnt)
         return self
 
+    def suggest_similarity_indexes(self) -> Self:
```
ludwiktrammer, line 147: Why is such flag-setter method preferable to this being just a construction argument?

src/dbally/views/freeform/text2sql/_autodiscovery.py
```diff
@@ -295,7 +306,11 @@ async def _suggest_similarity_indexes(
                 template=similarity_template,
                 fmt={"table_summary": description, "column_name": column.name, "values": example_values},
             )
-            similarity[column_name] = similarity_type
+
+            similarity_type = similarity_type.upper()
+
+            if similarity_type in ["SEMANTIC", "TRIGRAM"]:
```
ludwiktrammer, line 312: This should probably use values from the `Text2SQLSimilarityType` enum instead.

src/dbally/views/freeform/text2sql/_autodiscovery.py
mhordynski: Just to stick to the fluent api pattern that is used while configuring autodiscovery, also it may be possible in the future to have some extra configuration on this feature

src/dbally/views/freeform/text2sql/_config.py
```diff
@@ -1,10 +1,20 @@
 from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
-from typing import Dict, Optional
+from typing import Dict, Iterable, Optional, Tuple
 
 import yaml
 
 
+class Text2SQLSimilarityType(str, Enum):
```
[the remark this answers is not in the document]
mhordynski: this approach with `(str, Enum)` should be fully compatible with higher python versions. I'm uncertain about the benefits we would gain by using this 3rd party lib here.

src/dbally/collection.py
mhordynski: Yeah, you are right. I think we should have more generic structure for localization info (view / method / argument doesn't always make sense)

PR #28, opened 2024-05-10
refactor: refactor LLMOptions

micpst:
This PR adds default LLM options for LLM client and enables custom LLM parameterization in `collection.ask`.


src/dbally/collection.py
```diff
@@ -208,12 +221,18 @@ async def ask(self, question: str, dry_run: bool = False, return_natural_respons
...
             n_retries=self.n_retries,
             dry_run=dry_run,
+            llm_options=llm_options,
         )
         end_time_view = time.monotonic()
 
         textual_response = None
         if not dry_run and return_natural_response:
-            textual_response = await self._nl_responder.generate_response(view_result, question, event_tracker)
+            textual_response = await self._nl_responder.generate_response(
```
ds-sebastianchwilczynski, line 230: Maybe this is overthinking but I imagine such a scenario, in which I want to use different LLM for View selector, collection and nl_responder. To exemplify nl_responder gets a lot of input tokens but doesn't have to perform complicated "thinking" skills, so I can use smaller LLM and save some money. 

Now the question is whether such an approach will not lead to problems when different clients are used within different blocks.

src/dbally/data_models/llm_options.py
```diff
@@ -9,13 +11,29 @@ class LLMOptions:
...
+    n: Union[Optional[int], NotGiven] = NOT_GIVEN
+    presence_penalty: Union[Optional[float], NotGiven] = NOT_GIVEN
+    seed: Union[Optional[int], NotGiven] = NOT_GIVEN
+    stop: Union[Optional[Union[str, List[str]]], NotGiven] = NOT_GIVEN
+    temperature: Union[Optional[float], NotGiven] = NOT_GIVEN
+    top_p: Union[Optional[float], NotGiven] = NOT_GIVEN
 
     dict = asdict
+
+    def __or__(self, other: "LLMOptions") -> "LLMOptions":
```
ds-sebastianchwilczynski, line 25: Can we have a test for it? Looks good but if there is an issue with this, finding such a bug can be not so funny

src/dbally/data_models/llm_options.py
```diff
@@ -1,6 +1,8 @@
 from dataclasses import asdict, dataclass
 from typing import List, Optional, Union
 
+from openai import NOT_GIVEN, NotGiven
```
ds-sebastianchwilczynski, line 4: By doing this we force the user to download openai module even though they want to use different LLM provider. 

Maybe we should use define these types by ourselves? https://github.com/openai/openai-python/blob/main/src/openai/_types.py#L107

src/dbally/data_models/llm_options.py
```diff
@@ -1,6 +1,8 @@
...
+
 
 @dataclass
 class LLMOptions:
```
ds-sebastianchwilczynski, line 8: What is the idea to handle options for different LLMClients when they do not support some parameter?

ds-sebastianchwilczynski (changes requested):
Please write a test (for me can be an integration one that if we execute collection.ask() all LLM Options propagate correctly. + test the __or__ operator

Also the issue with openai dependency is important.

Rest for me is optional, I'll accept the PR if without them

src/dbally/collection.py
micpst: I think this is over-thinking it at this stage, we don't have the logic for many different LLM clients yet. In any case, in the future we can just put more parameters to parameterize each step separately.

src/dbally/data_models/llm_options.py
micpst: We can't define a custom sentinel here because it can't be parsed by the OpenAI client, LLMOptions are tied to a specific LLM client, and each integration will require us to write separate options. This is related to the fact that each LLM client has a slightly different API and requires different parameters to be set, so we need this separation anyway.

src/dbally/data_models/llm_options.py
micpst: This `LLMOptions` is tied to the OpenAI client, as originally stated in the comments. Each new integration will have its own `LLMOption` class to support a specific API.

src/dbally/data_models/llm_options.py
micpst: I literally tried this earlier and received an error from the openai module.

ds-sebastianchwilczynski: A nasty solution would be to convert these inside OpenAIClient.call

src/dbally/data_models/llm_options.py
ds-sebastianchwilczynski: ahh okay

src/dbally/data_models/llm_options.py
micpst: I would leave it in the `dbally.data_models.llm_options` module and only rename the file and create a base class for next integrations. What do you think?

cc @mhordynski @ludwiktrammer

src/dbally/data_models/llm_options.py
micpst: The merge functionality has been removed.

tests/integration/test_llm_options.py
```diff
@@ -0,0 +1,85 @@
...
+        self.model_name = "gpt-4"
+        self.default_options = default_options
+        self._prompt_builder = Mock()
+
+    async def call(self, *_, **__) -> str:
+        ...
+
+
+@pytest.mark.asyncio
+async def test_llm_options_propation():
```
ds-sebastianchwilczynski, line 40: a typo: propagation

ds-sebastianchwilczynski (approved):

tests/integration/test_llm_options.py
micpst: Fixed, thanks for catching!

mhordynski (approved):
lgtm

src/dbally/llm_client/base.py
```diff
@@ -2,69 +2,79 @@
...
-            frequency_penalty=frequency_penalty,
-            max_tokens=max_tokens,
-            n=n,
-            presence_penalty=presence_penalty,
-            seed=seed,
-            stop=stop,
-            temperature=temperature,
-            top_p=top_p,
-        )
+        options = options if options else self.default_options
```
ludwiktrammer, line 64: I believe the idea was for the options to be _merged_, so one can overwrite individual options, while leaving the rest as-is.

src/dbally/collection.py
```diff
@@ -174,6 +180,7 @@ async def ask(self, question: str, dry_run: bool = False, return_natural_respons
             dry_run: if True, only generate the query without executing it
             return_natural_response: if True (and dry_run is False as natural response requires query results),
                                      the natural response will be included in the answer
+            llm_options: options to use for the LLM client.
```
ludwiktrammer, line 183: "options to use for the LLM client" adds almost no additional information beyond the name of the argument itself. I think it would be worth having a more detailed information, including how those options will interact with the LLM client's default options.

src/dbally/llm_client/base.py
micpst: Yes, I know, this was the case with the openai and anthropic integration, where they explicitly provide a sentinel for the `NotGiven` option, which helps in the implementation of merging. The problem is with the future integration with vertexai, which does not have a `NotGiven` sentinel, all parameters are optional and there is no way to distinguish if the user wanted to change a parameter or not, we would have to define a custom `NotGiven` sentinel and write a custom merge function for it. Therefore, I decided to override the options, I think this solution is cleaner, the user can easily modify default `LLMOptions` before passing them to `collection.ask` getting a similar effect.

src/dbally/collection.py
micpst: I will correct this in PR with anthropic integration.

src/dbally/llm_client/base.py
ludwiktrammer: Ok, that makes sense, although in my mind it might also make sense to have our own provided-independent `NotGiven` that would be replaced with a provider-specific value at the very end (without exposing the provider-specific values to our end users)

micpst: Good idea, added in #29.

PR #29, opened 2024-05-14
refactor: add LLMOptions merge

micpst:
This PR adds merge support for LLMOptions as requested in #28

### DIFF
--- a/src/dbally/llm_client/base.py
+++ b/src/dbally/llm_client/base.py
@@ -4,12 +4,15 @@
 import abc
 from abc import ABC
 from dataclasses import asdict, dataclass
-from typing import Dict, Generic, Optional, Type, TypeVar, Union
+from typing import Any, ClassVar, Dict, Generic, Optional, Type, TypeVar, Union
 
 from dbally.audit.event_tracker import EventTracker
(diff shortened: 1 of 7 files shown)


ludwiktrammer (approved):

src/dbally/llm_client/base.py
```diff
@@ -19,7 +22,37 @@ class LLMOptions(ABC):
...
+            else self_dict[key]
+            for key in self_dict
+        }
+
+        return self.__class__(**updated_dict)
+
+    def dict(self) -> Dict[str, Any]:
+        """
+        Creates a dictionary representation of the LLMOptions instance.
+        If a value is None, it will be replaced with the _not_given value.
```
ludwiktrammer, line 46: I'm not a fun of mentioning private members in public documentation - I think either it should be public (if it is of interest to the end users) or not mentioned at all. So maybe just "replaced with a provider-specific not-given sentinel"?

micpst: You're right, thanks for catching

micpst: You're right, thanks for catching

PR #30, opened 2024-05-14
refactor: refactor prompt builder

micpst:
This PR removes hard-coded `PromptBuilder` in `LLMClient` and fixes issues with prompt builder for new LLM models integrations.

### DIFF
--- a/src/dbally/prompts/prompt_builder.py
+++ b/src/dbally/prompts/prompt_builder.py
@@ -13,19 +13,19 @@ class PromptBuilder:
     def __init__(self, model_name: Optional[str] = None) -> None:
         """
         Args:
-            model_name: Name of the model to load a tokenizer for.
-                        Tokenizer is used to append special tokens to the prompt. If empty, no tokens will be added.
+            model_name: name of the tokenizer model to use. If provided, the tokenizer will convert the prompt to the
+                format expected by the model. The model_name should be a model available on huggingface.co/models.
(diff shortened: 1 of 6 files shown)


src/dbally/prompts/prompt_builder.py
```diff
@@ -13,19 +13,19 @@ class PromptBuilder:
     def __init__(self, model_name: Optional[str] = None) -> None:
         """
         Args:
-            model_name: Name of the model to load a tokenizer for.
-                        Tokenizer is used to append special tokens to the prompt. If empty, no tokens will be added.
+            model_name: name of the tokenizer model to use. If provided, the tokenizer will convert the prompt to the
```
ludwiktrammer, line 16: nit (non-blocking): maybe the name `hg_model_name` would make the fact that this is only for Hugging Face models (and one shouldn't put the names of other models here) more explicit? Or maybe even we should have a separate HuggingFacePromptBuilder (probably inheriting from `PromptBuilder`) so the general `PromptBuilder` doesn't containg HG-specific stuff?

ludwiktrammer (approved):

src/dbally/prompts/prompt_builder.py
micpst: I like the idea of a separate class. We don't use Hugging Face stuff yet, but we could keep it for future integrations.

PR #31, opened 2024-05-14
feat(llm): add Anthropic LLM client

micpst:
This PR adds support for Anthropic's Claude models.


src/dbally/_exceptions.py
```diff
@@ -0,0 +1,42 @@
+class DballyError(Exception):
```
mhordynski, line 1:
```suggestion
class DbAllyError(Exception):
```

src/dbally/_exceptions.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Base class for all exceptions raised by dbally.
```
mhordynski, line 3:
```suggestion
    Base class for all exceptions raised by db-ally.
```

src/dbally/nl_responder/token_counters.py
```diff
@@ -34,6 +34,33 @@ def count_tokens_for_openai(messages: ChatFormat, fmt: Dict[str, str], model: st
...
+
+    Returns:
+        Number of tokens in the messages.
+
+    Raises:
+        ImportError: If anthropic package is not installed.
+    """
+
+    try:
+        from anthropic._tokenizers import sync_get_tokenizer  # pylint: disable=import-outside-toplevel
```
mhordynski, line 53: Is there async version? Maybe we should use it instead

src/dbally/prompts/common_validation_utils.py
```diff
@@ -47,3 +47,23 @@ def check_prompt_variables(chat: ChatFormat, variables_to_check: Set[str]) -> Ch
             "You need to format the following variables: {variables_to_check}"
         )
     return chat
+
+
+def extract_system_prompt(chat: ChatFormat) -> Tuple[ChatFormat, str]:
```
mhordynski, line 52: Maybe it would be more convienent as a method on ChatFormat class?

micpst: `ChatFormat` is not a class, it's just an type alias for tuple here.

src/dbally/nl_responder/token_counters.py
micpst: Yes, there is. Good idea, will change it.

src/dbally/llm_client/anthropic.py
```diff
@@ -0,0 +1,122 @@
...
+        response_format: Optional[Dict[str, str]],
+        options: AnthropicOptions,
+        event: LLMEvent,
+    ) -> str:
+        """
+        Calls the Anthropic API endpoint.
+
+        Args:
+            prompt: Prompt as an Anthropic client style list.
+            response_format: Optional argument used in the OpenAI API - used to force the json output.
```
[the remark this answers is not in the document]
micpst: I was thinking of moving `response_format` argument to `LLMOptions`, which would make the interface cleaner. What do you think?

cc @mhordynski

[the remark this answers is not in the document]
ludwiktrammer: I think `response_format` should remain coupled with the prompt (i.e., be part of the PromptTemplate class), because whether we expect json fully depend on the prompt itself.

PR #32, opened 2024-05-15
fix: fix AttributeError in LangSmithEventHandler

micpst:
This PR adds an optional API key for the LangSmith API and fixes a `AttributeError` in `request_end()` when using `LangSmithEventHandler`.


mhordynski (approved):
lgtm

PR #34, opened 2024-05-16
feat: Add elastic store

karllu3:
Add elastic store for similarity search.


src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Initializes the ElasticStore.
+
+        Args:
+            index_name: The name of the index.
+            embedding_client: The client to use for creating text embeddings.
+
+        """
+        super().__init__()
+        self.es = Elasticsearch(
```
mhordynski, line 34: Can we use asynchronous client here?

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+                - "column" (str): The original word.
+                - "search_vector" (numpy.ndarray): The embedding vector for the word, as a flat 1D array.
+        """
+        for word in data:
+            embedding = np.array(await self.embedding_client.get_embeddings([word]), dtype=np.float32).reshape(-1)
+            yield {"_index": self.index_name, "column": word, "search_vector": embedding}
+
+    async def store(self, data: List[str]) -> None:
+        """
+        Stores the data in a faiss index on disk.
```
mhordynski, line 72: docstring outdated

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @staticmethod
+    def _filter_response(res):
+        """Extracts and returns a specific field from the first hit in an Elasticsearch response.
+
+        This function processes a response from an Elasticsearch query to extract a specific nested field ('column')
+        from the first element in the list of hits, if any exist. If there are no hits, it returns None.
+
+        Args:
+            res (dict): The response dictionary from an Elasticsearch query.
```
mhordynski, line 101: Let's use typings as annotation in method signarute rather than in docstring

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
+# How-To: Use Elastic to Store Similarity Index
+
+[The similarity index](../concepts/similarity_indexes.md) is a feature provided by db-ally designed to map user input to the closest matching value in a data source. In this guide, we will show you how to execute similarity search by using elastic search.
```
mhordynski, line 3: Let's focus here on the ElasticStore class:

1. Add ElasticStore to `docs/reference/similarity/similarity_store/elastic.md`
2. Here we can write that `ElasticStore[link to reference]` can be used as a store in `SimilarityIndex[link to similarity index concept]` in similiar manner that we have in chroma doc

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+
+* Store specific (docker container creation printed) output to following variables in the project .venv file:
+  - ELASTIC_STORE_CONNECTION_STRING (e.g ```https://localhost:9200```),
+  - ELASTIC_CERT_PATH (e.g ```/home/username/certs/http_ca.crt```)
+  - ELASTIC_AUTH_USER (e.g ```"elastic```)
+  - ELASTIC_USER_PASSWORD (e.g ```uazxUIzCHwA7hcN6+7-5```)
+
+* Install elastic search python client
+```commandline
+pip install elasticsearch==8.13.1
```
mhordynski, line 27: Let's add extras_require in setup.cfg file to let our users install it with `pip install dbally[elasticsearch]`

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+## Implementing a SimilarityIndex
+
+To implement a Similarity Index with elastic store create ElasticStore object and pass it to a store argument.
+In the following example sensitive data has been loaded from .env file.
+
+```python
+load_dotenv()
+
+country_similarity = SimilarityIndex(
+    fetcher=SimpleSqlAlchemyFetcher(
```
mhordynski, line 39: I believe we should use a dummy fetcher, similar to the example in the ChromadbStore documentation. This way, we won't need to set up a database to run this example, making it more standalone.

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+        http_auth_tuple=(os.environ["ELASTIC_AUTH_USER"], os.environ["ELASTIC_USER_PASSWORD"]),
+        embedding_client=OpenAiEmbeddingClient(
+            api_key=os.environ["OPENAI_API_KEY"],
+        ),
+    ),
+)
+```
+
+You can then use this store to create a similarity index that maps user input to the closest matching value.
+For example To Search 'Russia' while querying 'RUS' or 'United States` while asking about the USA.
```
mhordynski, line 57:
```suggestion
Now you can use this index to map user input to the closest matching value. For example, a user may type 'United States' and our index would return 'USA'.```

Comment (mhordynski) [reply docs/how-to/use_elastic_store.md:57]: Also we may add code snippet with running it (running `.update()` method and after that `.similar()`)

File: src/dbally/similarity/elastic_store.py [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+
+    async def generate_data(self, data):
+        """Asynchronously generates and yields documents with embeddings for a list of words.
+
+        This coroutine iterates over a list of strings, fetches their embeddings using an asynchronous client,
+        and yields documents formatted for indexing in Elasticsearch. Each document contains the word, and
+        its corresponding embedding, reshaped as a flat array.
+
+        Args:
+            data (list of str): A list of words for which embeddings are to be generated.
```
mhordynski, line 58: Let's move typing to method signature

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+docker network create elastic
+docker pull docker.elastic.co/elasticsearch/elasticsearch:8.13.4
+docker run --name es01 --net elastic -p 9200:9200 -it -m 1GB docker.elastic.co/elasticsearch/elasticsearch:8.13.4
+# Copy the generated elastic password and enrollment token. These credentials are only shown when you start Elasticsearch for the first time. You can regenerate the credentials using the following commands.
+docker cp es01:/usr/share/elasticsearch/config/certs/http_ca.crt .
+curl --cacert http_ca.crt -u elastic:$ELASTIC_PASSWORD https://localhost:9200
+```
+
+* Store specific (docker container creation printed) output to following variables in the project .venv file:
+  - ELASTIC_STORE_CONNECTION_STRING (e.g ```https://localhost:9200```),
```
micpst, line 20: These envs don't render well in the documentation, they are mixed up with the steps and it's unreadable. Could you list them one by one, like this A, B, C? You don't need to give example values here.

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+```python
+load_dotenv()
+
+country_similarity = SimilarityIndex(
+    fetcher=SimpleSqlAlchemyFetcher(
+        engine,
+        table=Candidate,
+        column=Candidate.country,
+    ),
+    store=ElasticStore(
```
micpst, line 44: Could you initialize each component separately and add a small description to it, it's really hard to read here.

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,61 @@
...
+pip install elasticsearch==8.13.1
+```
+
+## Implementing a SimilarityIndex
+
+To implement a Similarity Index with elastic store create ElasticStore object and pass it to a store argument.
+In the following example sensitive data has been loaded from .env file.
+
+```python
+load_dotenv()
```
micpst, line 36: I would remove the `load_dotenv()` call here, as it isn't related to the similarity index API showcase.

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+from elasticsearch import AsyncElasticsearch
+from elasticsearch.helpers import async_bulk
+
+from dbally.embedding_client.base import EmbeddingClient
+from dbally.similarity.store import SimilarityStore
+
+
+class ElasticStore(SimilarityStore):
+    """
+    The ElasticStore class stores text embeddings using knnSearch.
```
micpst, line 13: Is it really knnSearch? What if the user specifies a different algorithm when initializing the class?

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+
+    def __init__(
+        self,
+        index_name: str,
+        embedding_client: EmbeddingClient,
+        host: str,
+        http_auth_tuple: Tuple[str, str],
```
micpst, line 21: Break this tuple into two separate arguments - login and password

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+        ca_cert_path: str,
+        search_algorith: Optional[Dict] = None,
+    ) -> None:
+        """
+        Initializes the ElasticStore.
+
+        Args:
```
micpst, line 28: Describe all init params  in docs string please

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+            index_name: The name of the index.
+            embedding_client: The client to use for creating text embeddings.
+
+        """
+        super().__init__()
+        self.es = AsyncElasticsearch(
```
micpst, line 34: Don’t use abbrs, please rename to `self.client`

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+            "properties": {
+                "search_vector": {
+                    "type": "dense_vector",
+                    "index": "true",
+                    "similarity": "cosine",
+                }
+            }
+        }
+
+        await self.es.indices.delete(index=self.index_name)
```
micpst, line 88: Do we have to drop whole index here?

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        Finds the most similar text in the store or returns None if no similar text is found.
+
+        Args:
+            text: The text to find similar to.
+
+        Returns:
+            The most similar text or None if no similar text is found.
+        """
+        embedding = np.array(await self.embedding_client.get_embeddings([text]), dtype=np.float32).reshape(-1)
```
micpst, line 122: Do we really need this numpy call here? Numpy is FAISS dependency, afaik

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+
+        await self.es.indices.delete(index=self.index_name)
+        await self.es.indices.create(index=self.index_name, mappings=mappings)
+
+        await async_bulk(self.es, self.generate_data(data))
+
+    @staticmethod
+    def _filter_response(res):
```
micpst, line 94: This additional method is unnecessary, make it one-liner and call directly in `find_similar()` method

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+            The most similar text or None if no similar text is found.
+        """
+        embedding = np.array(await self.embedding_client.get_embeddings([text]), dtype=np.float32).reshape(-1)
+
+        for key in self.search_algorithm:
+            self.search_algorithm[key]["query_vector"] = embedding
+            break
+
+        search_results = await self.es.search(
+            **self.search_algorithm,
```
micpst, line 129: If we're going to use only knn, as stated in the class description, parameterize knn explicitly, don't unpack it

src/dbally/similarity/elastic_store.py
```diff
@@ -0,0 +1,132 @@
...
+        self.indices = []
+        self.search_algorithm = search_algorith or {
+            "knn": {
+                "field": "search_vector",
+                "k": 10,
+                "num_candidates": 50,
+            }
+        }
+
+    async def generate_data(self, data):
```
micpst, line 50: This method `generate_data()` seems unnecessary to me. It looks quite simple, and is described as something really complex. Create these actions inside the `store()` method and use list comprehension.

micpst (changes requested):

src/dbally/similarity/elastic_store.py
karllu3: Numpy is good /fast to convert list of strings to list of float which is required by elastic.

src/dbally/similarity/elastic_store.py
karllu3: As I play with elastic store it is the most safe way due to semi automatic mapping.

src/dbally/similarity/elastic_store.py
micpst: Yes, numpy is fast, but embedding client returns list of floats, so there is no need to cast it to numpy array.

src/dbally/similarity/elastic_store.py
micpst: Ok, that makes sense. You might want to consider adding the `ignore_unavailable=True` flag here to avoid errors on first run.

src/dbally/similarity/elastic_store.py
karllu3: You are right, the problem here was type List[List[float]] while elastic expected List[float]. Fixed

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,96 @@
...
+curl --cacert http_ca.crt -u elastic:$ELASTIC_PASSWORD https://localhost:9200
+
+# For vector search it is required to enroll to appropiate subscryption level or trial verison which supports machine learning.
+# It is required to download ELSER model for example through kibana
+docker run --name kib01 --net elastic -p 5601:5601 docker.elastic.co/kibana/kibana:8.13.4
+# [Description how to download/enable ELSER model](https://www.elastic.co/guide/en/machine-learning/current/ml-nlp-elser.html)
+
+```
+
+* Install elastic search python client
```
micpst, line 28:
```suggestion
* Install elasticsearch extension
```

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,96 @@
...
+```commandline
+pip install dbally[elasticsearch,litllm]
```
micpst, line 30:
```suggestion
pip install dbally[elasticsearch]
```

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,96 @@
...
+```python
+class DummyCountryFetcher(SimilarityFetcher):
+    async def fetch(self):
+        return ["United States", "Canada", "Mexico"]
+```
+
+### Data store:
+Elastic store similarity search works on embeddings. For create embeddings the embedding client is passed as an argument.
+For example, we can use:
+* [OpenAiEmbeddingClient][dbally.embedding_client.OpenAiEmbeddingClient].
```
micpst, line 48: This is outdated, use `LiteLLMEmbeddingClient`

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,96 @@
...
+```python
+from dbally.similarity import SimpleSqlAlchemyFetcher, SimilarityIndex
+
+country_similarity = SimilarityIndex(
+    fetcher=DummyCountryFetcher(),
+    store=data_store
+)
+```
+
+and [update it and find the closest matches in the same way as in built-in similarity indices](./use_custom_similarity_store.md/#using-the-similar
```
micpst, line 88: This link doesn't render

src/dbally/similarity/elastic_vector_search.py
```diff
@@ -0,0 +1,170 @@
...
+            ca_cert_path (str): The path to the CA certificate for SSL/TLS verification.
+        """
+        super().__init__()
+        self.client = AsyncElasticsearch(
+            hosts=host,
+            http_auth=(http_user, http_password),
+            ca_certs=ca_cert_path,
+        )
+        self.index_name = index_name
+        self.indices = []
```
micpst, line 41: We don't use it anywhere, we can get rid of it

src/dbally/similarity/elastic_vector_search.py
```diff
@@ -0,0 +1,170 @@
...
+        index_name: str,
+        host: str,
+        http_user: str,
+        http_password: str,
+        ca_cert_path: str,
+    ) -> None:
+        """
+        Initializes the Elastic Vector Store.
+
+        Args:
```
micpst, line 27: Please remove typing from doc string

src/dbally/similarity/elastic_vector_search.py
```diff
@@ -0,0 +1,170 @@
...
+        - Creates a new trained model with the specified model ID.
+        - Polls the model definition status until the model is fully defined.
+        - Starts the deployment of the model with a specified number of allocations.
+        - Polls the deployment status until the model deployment is started.
+        - Sets up an ingest pipeline using the deployed model.
+
+        """
+        try:
+            await self.client.ml.delete_trained_model(model_id=".elser_model_2", force=True)
+            print("Model deleted successfully, We will proceed with creating one")
```
micpst, line 58: Maybe we should use `EventTracker` here instead of raw prints? cc @mhordynski

docs/how-to/use_elastic_store.md
```diff
@@ -0,0 +1,96 @@
...
+For example, we can use:
+* [OpenAiEmbeddingClient][dbally.embedding_client.OpenAiEmbeddingClient].
+
+```python
+from dbally.embeddings.litellm import LiteLLMEmbeddingClient
+
+embeddings=LiteLLMEmbeddingClient((api_key="your-api-key")
+```
+
+To implement a Similarity Index with elastic store create ElasticStore object and pass it to a store following argument:
```
micpst, line 56: This doesn't render in the docs, please reformat this. Also, I'm not sure if we should copy paste these argument definitions from the doc string

micpst:
Please check the docs again and make sure they're consistent with what we already have, and before the next review, please build it and check that everything renders correctly. Thanks!

src/dbally/similarity/elastic_vector_search.py
```diff
@@ -0,0 +1,170 @@
...
+        1. Deploys the ELSER model by calling `deploy_elser_mode`.
+        2. Defines the mappings for the Elasticsearch index.
+        3. Deletes the existing index if it exists.
+        4. Creates a new index with the specified mappings and settings.
+        5. Stores the provided data in the Elasticsearch index using the default ingest pipeline.
+
+        Args:
+            data: The data to store in the Elasticsearch index.
+        """
+        await self.deploy_elser_model()
```
mhordynski, line 114: Why we are deploying elser model? I think it should be responsibility of the user - maybe even user already has model deployed, or has some other orchiestration to do this.

src/dbally/similarity/elastic_vector_search.py
```diff
@@ -0,0 +1,110 @@
...
+            http_auth=(http_user, http_password),
+            ca_certs=ca_cert_path,
+        )
+        self.index_name = index_name
+
+    async def store(self, data: List[str]) -> None:
+        """
+        Stores the given data in an Elasticsearch store.
+
+        This function performs the following steps:
```
micpst, line 46: Please update this doc, we no longer deploy ELSER model here.

src/dbally/similarity/elasticsearch_store.py
```diff
@@ -0,0 +1,109 @@
...
+        embedding_client: EmbeddingClient,
+        host: str,
+        http_user: str,
+        http_password: str,
+        ca_cert_path: str,
+    ) -> None:
+        """
+        Initializes the ElasticsearchStore.
+
+        Args:
```
micpst, line 29: Please remove argument typing here as well for consistency.

micpst (approved):

PR #35, opened 2024-05-16
feat(llm): integrate LLMClient with litellm

micpst:
This PR adds support for more than 100+ LLMs by replacing the openai package in `LLMClient` with the litellm package.


micpst:
If we accept the transition to `litellm`, I will update the docs in the next PR.

setup.cfg
```diff
@@ -41,19 +41,18 @@ install_requires =
     numpy>=1.24.0
 
 [options.extras_require]
-openai =
```
mhordynski, line 44: openai is still required for openai embeddings

mhordynski (approved):
@micpst good job :muscle:

PR #38, opened 2024-05-22
docs: update guides and API refs

micpst:
This PR  updates guides and API references for `llms` and `embeddings` modules.


docs/how-to/llms/custom.md
```diff
@@ -0,0 +1,118 @@
...
+
+class MyLLM(LLM[LiteLLMOptions]):
+    _options_cls = LiteLLMOptions
+```
+
+In this example we will be using `LiteLLMOptions`, which contain all options supported by most popular LLM APIs. If you need a different interface, see [Customising LLM Options](#customising-llm-options) to learn how to implement it.
+
+### Step 2: Create the custom LLM client
+
+The `_client` property is an abstract method that must be implemented in your subclass. This property should return an instance of `LLMClient` that your LLM will use to interact with the model:
```
mhordynski, line 27: I think it would be nice to link the API Reference page in every reference like `LLMClient` or `LiteLLMOptions`

mhordynski (approved):
lgtm

PR #39, opened 2024-05-28
draft: feat: Gradio Adater implementation

karllu3:


setup.cfg
```diff
@@ -39,6 +39,7 @@ install_requires =
     tabulate>=0.9.0
     click~=8.1.7
     numpy>=1.24.0
+    gradio>=4.31.5
```
mhordynski, line 42: gradio is not needed in main requirements

src/dbally/utils/gradio_adapter.py
```diff
@@ -0,0 +1,173 @@
...
+
+from dbally import BaseStructuredView
+from dbally.collection import Collection
+from dbally.prompts import PromptTemplateError
+from dbally.similarity import SimilarityIndex
+from dbally.utils.errors import NoViewFoundError, UnsupportedQueryError
+from dbally.utils.log_to_file import FileLogger
+
+CONSOLE_FILE_NAME = "console.log"
+sys.stdout = FileLogger(CONSOLE_FILE_NAME)
```
mhordynski, line 15: we shouldn't override sys.stdout. I think it would be nicer to:

* implement custom parameter to CLIEventHandler to specify where output should be written
* inject dedicated instance of this event_handler into collection to handle gradio output

src/dbally/utils/gradio_adapter.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def __init__(self, preview_limit: int = 20):
+        """
+        Initializes the GradioAdapter with a preview limit.
+
+        Args:
+            preview_limit: The maximum number of preview data records to display. Default is 20.
+        """
+        self.preview_limit = preview_limit
+        self.similarity_store_list = []
```
mhordynski, line 31: we don't need to store any information about similarity stores. That is already handled by Collection. Also updating similarity stores should not happen inside gradio adapter.

src/dbally/utils/gradio_adapter.py
```diff
@@ -0,0 +1,173 @@
...
+        self.collection = None
+        sys.stdout.flush()
+
+    async def ui_load_preview_data(self, selected_view_name: str) -> Tuple[pd.DataFrame, str, None, None, None, None]:
```
mhordynski, line 35: I think this method may be private. As well as `ui_ask_query` and `load_preview_data`

mhordynski (changes requested):

src/dbally/utils/gradio_adapter.py
micpst, line 1: We need to refactor the `utils` module to have a common reusable components. Let's move it outside of `utils` to a separate `visualizer.py` file or create a `visualizer.gradio` module.

mhordynski: I think it should be something like:
`from dbally.gradio import GradioAdapter`

mhordynski: and maybe even just have a function for that exposed rather than class, I think it would be nicer for user with API like this:

```
from dbally.gradio import create_gradio_interface

create_gradio_interface(collection).launch()
```

micpst: `from dbally.gradio import create_gradio_interface` looks good, we can go for it.

docs/how-to/visualize_views.md
```diff
@@ -0,0 +1,37 @@
...
+Launch the gradio interface. To publish public interface pass argument `share=True`
+```python
+    gradio_interface.launch()
+```
+
+The endpoint is set by by triggering python module with Gradio Adapter launch command.
+Private endpoint is set to http://127.0.0.1:7860/ by default.
+
+## Links
+* [Example Gradio Interface](../../examples/visualize_views_code.py)
```
micpst, line 37: This link doesn't work, please replace it with a link to the file in github repo.

docs/how-to/visualize_views.md
```diff
@@ -0,0 +1,37 @@
...
+## Installation
+```bash
+pip install dbally[gradio]
+```
+
+## Create own gradio interface
+Define collection with implemented views
+
+```python
+    llm = LiteLLM(model_name="gpt-3.5-turbo")
```
micpst, line 15: Please fix indentation in python code samples

examples/recruting/candidate_view_with_similarity_store.py
```diff
@@ -0,0 +1,70 @@
...
+Candidate = Base.classes.candidates
+
+country_similarity = SimilarityIndex(
+    fetcher=SimpleSqlAlchemyFetcher(
+        engine,
+        table=Candidate,
+        column=Candidate.country,
+    ),
+    store=FaissStore(
+        index_dir="./similarity_indexes",
```
micpst, line 27: Faiss store crashes here.

examples/recruting/candidate_view_with_similarity_store.py
```diff
@@ -0,0 +1,70 @@
...
+        index_name="country_similarity",
+        embedding_client=LiteLLMEmbeddingClient(
+            model="text-embedding-3-small",  # to use openai embedding model
+            api_key=os.environ["OPENAI_API_KEY"],
```
micpst, line 31: We can get rid of this `api_key` here.

docs/quickstart/index.md
```diff
@@ -30,7 +30,7 @@ pip install dbally[litellm]
 
 ## Database Configuration
 
-In this guide, we will use an example SQLAlchemy database containing a single table named `candidates`. This table includes columns such as `id`, `name`, `country`, `years_of_experience`, `position`, `university`, `skills`, and `tags`. You can download the example database from [candidates.db](candidates.db). Alternatively, you can use your own database and models.
+In this guide, we will use an example SQLAlchemy database containing a single table named `candidates`. This table includes columns such as `id`, `name`, `country`, `years_of_experience`, `position`, `university`, `skills`, and `tags`. You can download the example database from [candidates.db](../../examples/recruting/candidates.db). Alternatively, you can use your own database and models.
```
micpst, line 33: This link doesn't work either.

src/dbally/gradio/gradio_interface.py
```diff
@@ -0,0 +1,181 @@
...
+        self.collection = user_collection
+        self.collection.add_event_handler(CLIEventHandler(self.log))
+
+        view_list = [*user_collection.list()]
+        if view_list:
+            default_selected_view_name = view_list[0]
+        else:
+            raise ValueError("No view to display")
+
+        if not view_list:
```
micpst, line 137: Above you raise an exception if `view_list` is empty, so it's not needed here. Maybe we shouldn't really do anything if there is no view to display, just render an empty boilerplate.

src/dbally/gradio/gradio_interface.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Raises:
+        ValueError: occurs when there is no view define in collection.
+    """
+    adapter = _GradioAdapter()
+    gradio_interface = await adapter.create_interface(user_collection, preview_limit)
+    return gradio_interface
+
+
+class _GradioAdapter:
```
micpst, line 32: Make it public

examples/visualize_views_code.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+async def main():
+    await country_similarity.update()
+    llm = LiteLLM(model_name="gpt-3.5-turbo")
+    collection = dbally.create_collection("candidates", llm, event_handlers=[CLIEventHandler()])
+    collection.add(CandidateView, lambda: CandidateView(engine))
+    collection.add(SampleText2SQLViewCyphers, lambda: SampleText2SQLViewCyphers(create_freeform_memory_engine()))
+    gradio_interface = await create_gradio_interface(user_collection=collection)
+    if gradio_interface:
```
micpst, line 20: `gradio_interface` is no longer optional

docs/how-to/visualize_views.md
```diff
@@ -0,0 +1,42 @@
...
+```python
+if gradio_interface:
+    gradio_interface.launch()
+```
+
+The endpoint is set by triggering python module with Gradio Adapter launch command.
+Private endpoint is set to http://127.0.0.1:7860/ by default.
+
+## Links
+* [Example Gradio Interface](https://github.com/deepsense-ai/db-ally/blob/lk/gradio_adapter/examples/visualize_views_code.py)
```
micpst, line 42: Reference main branch here

src/dbally/gradio/gradio_interface.py
```diff
@@ -0,0 +1,174 @@
...
+
+async def create_gradio_interface(user_collection: Collection, preview_limit: int = 20) -> Optional[gradio.Interface]:
+    """Adapt and integrate data collection and query execution with Gradio interface components.
+
+    Args:
+        user_collection: The user's collection to interact with.
+        preview_limit: The maximum number of preview data records to display. Default is 20.
+
+    Returns:
+        The created Gradio interface, or None if no data is available to load.
```
micpst, line 22: This is no longer optional, please also update the method signature.

examples/visualize_views_code.py
```diff
@@ -0,0 +1,24 @@
...
+import asyncio
+
+from recruiting.candidate_view_with_similarity_store import CandidateView, country_similarity, engine
+from recruiting.cypher_text2sql_view import SampleText2SQLViewCyphers, create_freeform_memory_engine
+
+import dbally
+from dbally.audit import CLIEventHandler
+from dbally.gradio import create_gradio_interface
+from dbally.llms.litellm import LiteLLM
+
```
PatrykWyzgowski, line 11: My suggestion is to make this script run independently. Otherwise please mention or remind the need for particular environment variables as their use is not seen here explicitly making the example fail.
```suggestion
from dotenv import load_dotenv
load_dotenv()
```

micpst: You are right, but loading dotenv does not make it error-proof. There is a self explanatory error if env is missing. We can also add an additional hint in the documentation to set this variable before running the script.

micpst (approved):

mhordynski (approved):
good job :medal_sports:

PR #40, opened 2024-05-29
feat(codegen): add text2sql views code generation

micpst:
This PR adds code generation for text2sql views and refactors autodiscovery for codegen.


src/dbally_cli/text2sql.py
```diff
@@ -0,0 +1,133 @@
...
+    Args:
+        file_path: The path to the file where the view will be generated.
+        db_url: The database connection string.
+        llm_object: The path to the LLM object.
+        llm_description: Generate a description using the LLM object.
+        similarity_index_factory: The path to the similarity index factory.
+    """
+    click.echo("Generating Text2SQL view...")
+
+    root, ext = os.path.splitext(file_path)
```
karllu3, line 64: better use pathlib it has been added in python 3.4

micpst: A matter of preference, it does not change the functionality.

mhordynski:
Code looks great :1st_place_medal: Let's add documentation and we can merge it

micpst:
I'll update the documentation in a separate PR for both code generation and freeform views.

mhordynski (approved):

PR #42, opened 2024-06-05
feat: few-shot selector

sgnatonski:
feat: added few shot selector to inject examples into prompt, added prompt input formatters abstraction


examples/recruting/views.py
```diff
@@ -57,3 +60,40 @@ def is_from_continent(  # pylint: disable=W0602, C0116, W9011
...
+    @decorators.view_filter()
+    def is_available_within_months(  # pylint: disable=W0602, C0116, W9011
+        self, months: int
+    ) -> sqlalchemy.ColumnElement:
+        start = date.today()
+        end = start + relativedelta(months=months)
+        return Candidate.available_from.between(start, end)
+
+    @decorators.few_shot()
+    def get_examples(self) -> List[FewShotExample]:  # pylint: disable=W0602, C0116, W9011
```
mhordynski, line 79: let's add `question: str` as an argument here - it will allow the user to select / rerank few-shot examples based on similarity

src/dbally/iql_generator/iql_prompt_template.py
```diff
@@ -60,3 +75,27 @@ def _validate_iql_response(llm_response: str) -> str:
...
+            "Please suggest which one(s) to call and how they should be joined with logic operators (AND, OR, NOT).\n"
+            "Remember! Don't give any comments, just the function calls.\n"
+            "The output will look like this:\n"
+            'filter1("arg1") AND (NOT filter2(120) OR filter3(True))\n'
+            "DO NOT INCLUDE arguments names in your response. Only the values.\n"
+            "You MUST use only these methods:\n"
+            "\n{filters}\n"
+            "It is VERY IMPORTANT not to use methods other than those listed above."
+            "Here's a list of examples:\n"
+            "\n{examples}\n"
```
mhordynski, line 94: hmm maybe we can put few-shots as a part of conversation, then we wouldn't have to create separate prompt for that.

for example:
```
    chat=(
        {
            "role": "system",
            "content": "You have access to API that lets you query a database:\n"
                            .......
        },
        {"role": "user", "content": "<QUESTION 1>"},
        {"role": "assistant", "content": "<ANSWER 1>"},
        {"role": "user", "content": "<QUESTION 2>"},
        {"role": "assistant", "content": "<ANSWER 2>"},
        {"role": "user", "content": "{question}"},
    ),
```

src/dbally/prompts/few_shot.py
```diff
@@ -0,0 +1,37 @@
...
+            answer_expr: it can be either a stringified expression or a lambda for greater safety and code completions.
+                in lambda case, it should be a single line expression (otherwise output correctness is not guaranteed)
+        """
+        self.question = question
+        self.answer_expr = answer_expr
+
+        if isinstance(self.answer_expr, str):
+            self.answer = self.answer_expr
+        else:
+            expr_source = textwrap.dedent(inspect.getsource(self.answer_expr))
```
micpst, line 25: This code in the `else:` block effectively cleans up and normalizes a piece of Python code. I think you could extract it into a separate method or at least briefly describe what it does in a doc string.

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,91 @@
...
+        filters: list of filters exposed by the view
+
+    Returns:
+        filters_for_prompt: filters formatted for prompt
+    """
+    filters_for_prompt = "\n".join([str(filter) for filter in filters])
+    return filters_for_prompt
+
+
+class AbstractInputFormatter:
```
micpst, line 26: I think it should be an abstract class, also don't use "Abstract" in the class name, just call it `InputFormatter`.

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,91 @@
...
+        question: str,
+    ) -> None:
+        self.filters = filters
+        self.question = question
+        self.examples = examples
+
+    def __call__(self, conversation_template: PromptTemplate) -> Tuple[PromptTemplate, Dict[str, str]]:
+        template_copy = copy.deepcopy(conversation_template)
+        sys_msg = template_copy.chat[0]
+        exisiting_msgs = [c for c in template_copy.chat[1:] if "is_example" not in c]
```
micpst, line 71: Please don't use single letters or abbrevs for variable names, just use full names.

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,91 @@
...
+        chat_examples = [
+            c
+            for e in self.examples
+            for c in [
+                {"role": "user", "content": e.question, "is_example": True},
+                {"role": "assistant", "content": e.answer, "is_example": True},
+            ]
+        ]
+        new_chat = (
```
micpst, line 80: This variable is unnecessary

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,91 @@
...
+    return filters_for_prompt
+
+
+class AbstractInputFormatter:
+    """
+    Formats provided parameters to a form acceptable by IQL prompt
+    """
+
+    @abstractmethod
+    def __call__(self, conversation_template: PromptTemplate) -> Dict[str, str]:
```
micpst, line 32: Please add doc string what this call does.

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,91 @@
...
+class DefaultInputFormatter(AbstractInputFormatter):
+    """
+    Formats provided parameters to a form acceptable by default IQL prompt
+    """
+
+    def __init__(self, filters: List[ExposedFunction], question: str) -> None:
+        self.filters = filters
+        self.question = question
+
+    def __call__(self, conversation_template: PromptTemplate) -> Tuple[PromptTemplate, Dict[str, str]]:
```
micpst, line 45: There is a discrepancy between the abstract class return type annotation and the implementation.

micpst:
Really good job with introducing this `InputFormatter` for `IQLGenerator`, we should also refactor `ViewSelector` and `NLResponder` to use it. Just wondering whether to add it in this PR. cc @mhordynski

mhordynski:
> Really good job with introducing this `InputFormatter` for `IQLGenerator`, we should also refactor `ViewSelector` and `NLResponder` to use it. Just wondering whether to add it in this PR. cc @mhordynski

Let's do this in separate PR

src/dbally/prompts/few_shot.py
```diff
@@ -0,0 +1,37 @@
...
+        Args:
+            question: sample question
+            answer_expr: it can be either a stringified expression or a lambda for greater safety and code completions.
+                in lambda case, it should be a single line expression (otherwise output correctness is not guaranteed)
+        """
+        self.question = question
+        self.answer_expr = answer_expr
+
+        if isinstance(self.answer_expr, str):
+            self.answer = self.answer_expr
```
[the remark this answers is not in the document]
sgnatonski: good point, although it depends a bit what we would like to fail first when the `answer_expr` cannot be parsed. if the logic is executed on `FewShotExample` creation and an exception happens we will immediately know that what we passed as a construction arguments is invalid, while having the logic in `cached_property` might give a false sense of no problems, because any exception would be thrown only when the property is accessed for the first time.
I'm fine with `cached_property` approach as it will certainly clean up initialization, so LMK if I should proceed and introduce `cached_property` :)

src/dbally/prompts/few_shot.py
sgnatonski: good point, I will fix that in next commit

src/dbally/prompts/input_format.py
sgnatonski: thank you, I agree

src/dbally/prompts/input_format.py
sgnatonski: you would like to assign directly to `template_copy.chat`, I presume?

src/dbally/prompts/input_format.py
sgnatonski: ah, thank you for spotting that

src/dbally/prompts/few_shot.py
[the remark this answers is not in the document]
mhordynski: We should leave that code in the `__init__` method to ensure that any syntax errors are caught as early as possible.

src/dbally/prompts/input_format.py
micpst: yes, exactly

src/dbally/prompts/few_shot.py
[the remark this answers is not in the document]
micpst: Yeah, you're right, let's leave it as it is.

src/dbally/prompts/input_format.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Formats filters for prompt
+
+    Args:
+        filters: list of filters exposed by the view
+
+    Returns:
+        filters_for_prompt: filters formatted for prompt
+    """
+    filters_for_prompt = "\n".join([str(filter) for filter in filters])
```
micpst, line 22: Unnecessary variable

src/dbally/views/freeform/text2sql/view.py
```diff
@@ -248,3 +262,15 @@ def list_similarity_indexes(self) -> Dict[AbstractSimilarityIndex, List[IndexLoc
                 if column.similarity_index:
                     indexes[column.similarity_index].append((self.__class__.__name__, table.name, column.name))
         return indexes
+
+    def examples(self, query: str) -> List[FewShotExample]:  # pylint: disable=unused-argument
```
micpst, line 266: I think this method should be moved to `dbally.views.base.BaseView` and renamed to `list_examples()`. Then implement it separately for structured and freeform views.

src/dbally/views/structured.py
```diff
@@ -56,14 +57,27 @@ async def ask(
...
+
         filter_list = self.list_filters()
+        few_shot_list = self.list_few_shot()
+        iql_generator = self.get_iql_generator(llm)
+
+        input_formatter = (
+            DefaultFewShotInputFormatter(
+                question=query,
+                filters=filter_list,
+                examples=[
```
micpst, line 69: Examples of structured views should be provided in a similar way to freeform views (related to refactoring `examples()` method).

micpst (approved):

PR #45, opened 2024-06-13
docs(views): add docs for freeform views

micpst:
This PR adds docs for freeform views.


docs/how-to/views/text-to-sql.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+The returned list contains [`TableConfig`](../../reference/views/text-to-sql.md#dbally.views.freeform.text2sql.config.TableConfig) objects for each table in the database that the LLM should know about, answering the question. At this point you can decide what kind of data the LLM model should have access to, whether at the table or column level, you can limit the visibility of the data at this point.
+
+!!! note
+    As you may notice there is no need to define IQL specific features like filters, filtering and building query will be handled be the LLM in this case.
+
+## Autodiscovery
+
+Db-ally allows you to automatically retrieve table configurations from any SQL database. You can also use `LLM` to fill in missing table descriptions, and suggest the use of a similarity index for a given column. To start the autodiscovery process, create an `AutodiscoveryBuilder` instance with the `configure_text2sql_auto_discovery` factory method, and then run the `discover` method to start the process.
```
mhordynski, line 45:
```suggestion
db-ally allows you to automatically retrieve table configurations from any SQL database. You can also use `LLM` to fill in missing table descriptions, and suggest the use of a similarity index for a given column. To start the autodiscovery process, create an `AutodiscoveryBuilder` instance with the `configure_text2sql_auto_discovery` factory method, and then run the `discover` method to start the process.
```

docs/how-to/views/text-to-sql.md
```diff
@@ -0,0 +1,110 @@
...
+from dbally.llms.litellm import LiteLLM
+from sqlalchemy import create_engine
+
+db = create_engine("sqlite://")
+llm = LiteLLM("gpt-3.5-turbo")
+
+collection = dbally.create_collection("text-to-sql", llm=llm)
+collection.add(CandidateView, lambda: CandidateView(db))
+
+response = await collection.ask("What is the meaning of life?")
```
mhordynski, line 109: I agree that it is v important question, but I'm not sure if db-ally can give you an answer

micpst: Yea, you're right 😅

PR #59, opened 2024-06-20
feat: collection hierarchy

karllu3:


mhordynski (changes requested):

src/dbally/audit/event_tracker.py
```diff
@@ -92,3 +92,15 @@ async def track_event(self, event: Event) -> AsyncIterator[EventSpan]:
             await handler.event_end(
                 span.data, event_context=contexts[handler], request_context=self._request_contexts[handler]
             )
+
+    async def log_message(self, message: str, log_level: LogLevel = LogLevel.INFO) -> None:
```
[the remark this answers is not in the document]
mhordynski: @micpst wdyt?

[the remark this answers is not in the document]
micpst: I like the idea of `FallbackEvent`. I have some doubts about passing `EventTracker` to execute the next collection in the sequence. This can cause confusion when two collections have different event handlers, as we'll overwrite the collection's handlers down the chain. Maybe we should think about creating a span here to log events from a given chain, and each subchain would have its own subspan to handle, which will end up with a nice trace. For different event handlers, we'll have just a span for a collection node, with enough information. This is a high-level idea, but I think it could work.

src/dbally/collection/collection.py
```diff
@@ -192,49 +226,69 @@ async def ask(
...
-        if len(views) == 1:
-            selected_view = next(iter(views))
-        else:
-            selected_view = await self._view_selector.select_view(
-                question=question,
-                views=views,
+        selected_view = None
+        result = None
+
+        try:
```
micpst, line 232: It's getting messy here. Maybe we should add a utils decorator to bind exceptions to handlers and move the chain execution to a separate method. I'm thinking of something like this:
```python
@handle_exception([NoViewFoundError,...], self.ask_fallback)
def ask(self, ...):
```

docs/concepts/collections.md
```diff
@@ -25,6 +25,31 @@ my_collection.ask("Find me Italian recipes for soups")
 
 In this scenario, the LLM first determines the most suitable view to address the query, and then that view is used to pull the relevant data.
 
+Sometimes, the selected view may not be able to answer the question and will raise an error. In such situations, the fallback collections can be used. This will cause a next view selection, but from the fallback collection.
```
ludwiktrammer, line 28: Just to bring my up to speed - why switching to a different collection when encountering errors in view execution? For me, the more intuitive solution would be to retry with the view, while providing the error information to the LLM. When I heard about "fallback collections" I thought they would be for cases when the view selector is not able to find any appropriate view in the origin collection, not for handling errors connected to individual views.

karllu3: We assume that llm pick wrong view for the ask therefore we want to look for another view. The idea is that in second collection would be freeform views that will operate as fallbacks.  When it will be in one collection just blacklisting checked first pick we are limited only to llm choice. In implemented case we can easily manage that second collection would have only freeform views.

ludwiktrammer: Ok. For me probably the jump from "error in a view" to "llm picked the wrong view" is non-obvious.

src/dbally/collection/collection.py
```diff
@@ -143,12 +182,154 @@ def list(self) -> Dict[str, str]:
...
+
+        fallback_event = FallbackEvent(
+            triggering_collection_name=self.name,
+            triggering_view_name=selected_view_name,
+            fallback_collection_name=self._fallback_collection.name,
+            error_description=repr(caught_exception),
+        )
+
+        async with event_tracker.track_event(fallback_event) as span:
+            result = await self._fallback_collection.ask(
```
mhordynski, line 317: `event_tracker` is missing in the call

src/dbally/collection/collection.py
```diff
@@ -143,12 +182,154 @@ def list(self) -> Dict[str, str]:
...
+        """
+
+        fallback_event = FallbackEvent(
+            triggering_collection_name=self.name,
+            triggering_view_name=selected_view_name,
+            fallback_collection_name=self._fallback_collection.name,
+            error_description=repr(caught_exception),
+        )
+
+        async with event_tracker.track_event(fallback_event) as span:
```
mhordynski, line 316: let's test how it looks in OTEL / Langsmith when we nest events inside, not sure if it is supported right now

examples/recruiting/views.py
```diff
@@ -75,7 +75,7 @@ def is_available_within_months(  # pylint: disable=W0602, C0116, W9011
         end = start + relativedelta(months=months)
         return Candidate.available_from.between(start, end)
 
-    def list_few_shots(self) -> List[FewShotExample]:  # pylint: disable=W9011
+    def list_few_shots(self) -> List[FewShotExample]:  # pylint: disable=W9011, C0116
```
PatrykWyzgowski, line 78: I think the docstring for this method is going to be short enough to put it here instead of disabling the linter.

mhordynski (approved):
lgtm

PR #62, opened 2024-06-24
feat: aggregations in structured views

PatrykWyzgowski:


src/dbally/views/structured.py
```diff
@@ -57,20 +57,32 @@ async def ask(
...
             filters=filters,
             examples=examples,
+            aggregations=[],
             event_tracker=event_tracker,
             llm_options=llm_options,
             n_retries=n_retries,
         )
         await self.apply_filters(iql)
 
+        iql = await iql_generator.generate_iql(
```
mhordynski, line 75: Here we overwrite `iql`, so the context later is lacking information about the filters

docs/how-to/views/custom.md
```diff
@@ -106,37 +106,41 @@ Let's implement the required `apply_filters` method in our `FilteredIterableBase
...
 
-    async def apply_filters(self, filters: IQLQuery) -> None:
-        """
-        Applies the selected filters to the view.
 
-        Args:
-            filters: IQLQuery object representing the filters to apply
-        """
-        self._filter = await self.build_filter_node(filters.root)
+super().__init__()
```
mhordynski, line 111: I think something went wrong with formatting here

src/dbally/iql_generator/prompt.py
```diff
@@ -85,3 +88,26 @@ def __init__(
...
+        {
+            "role": "system",
+            "content": "You have access to API that lets you query a database supporting SINGLE aggregation.\n"
+            "When prompted for just aggregation, use the following methods: \n"
+            "{aggregations}"
+            "DO NOT INCLUDE arguments names in your response. Only the values.\n"
+            "You MUST use only these methods:\n"
+            "\n{aggregations}\n"
+            "Structure output to resemble the following pattern:\n"
+            'aggregation1("Argument_in_lowercase")\n'
```
mhordynski, line 103: why in lowercase?

src/dbally/iql_generator/prompt.py
```diff
@@ -85,3 +88,26 @@ def __init__(
...
+            "It is VERY IMPORTANT not to use methods other than those listed above."
+            """If you DON'T KNOW HOW TO ANSWER DON'T SAY \"\" anything other than `UNSUPPORTED QUERY`"""
+            "This is CRUCIAL to put `UNSUPPORTED QUERY` text only, otherwise the system will crash. "
+            "Structure output to resemble the following pattern:\n"
+            'aggregation1("Argument_in_lowercase", another_base_python_datatype_argument)\n',
```
mhordynski, line 108: `another_base_python_datatype_argument` looks like iql could handle arbitrary python vars, rather than just inline constants.

src/dbally/views/structured.py
PatrykWyzgowski: Done.

src/dbally/iql_generator/prompt.py
PatrykWyzgowski: Fixed

src/dbally/iql_generator/prompt.py
PatrykWyzgowski: Adjusted.

docs/how-to/views/custom.md
PatrykWyzgowski: Fixed.

src/dbally/iql_generator/iql_generator.py
```diff
@@ -60,8 +67,12 @@ async def generate_iql(
         prompt_format = IQLGenerationPromptFormat(
             question=question,
             filters=filters,
+            aggregations=aggregations,
             examples=examples,
         )
+        if aggregations and not filters:
```
ludwiktrammer, line 73: So the same `IQLGenerator` class is used for two separate functionalities in the system with different prompts and the decision which one to execute depends on whether the "filters" attribute is present (and the list is non-empty) or not? That seems confusing to me and this behavior is not explained in the documentation of this class.

I believe those should be split into two separate classes. Or, at the very least, the switch of behavior should be very explicitly described in the docstring for the class itself and the "aggregations" and "filters" arguments.

src/dbally/iql_generator/prompt.py
```diff
@@ -53,10 +54,12 @@ def __init__(
             question: Question to be asked.
             filters: List of filters exposed by the view.
             examples: List of examples to be injected into the conversation.
+            aggregations: List of examples to be computed for the view (Currently 1 aggregation supported).
```
ludwiktrammer, line 57: The same comment as the one for `IQLGenerator` + "(Currently 1 aggregation supported)" sounds like you can only pass a single aggregation here, which as far as I understand is not true.

src/dbally/iql_generator/prompt.py
```diff
@@ -85,3 +88,24 @@ def __init__(
     ],
     response_parser=_validate_iql_response,
 )
+
+IQL_GENERATION_TEMPLATE_AGGREGATION = PromptTemplate[IQLGenerationPromptFormat](
+    chat=(
+        {
+            "role": "system",
+            "content": "You have access to API that lets you query a database supporting SINGLE aggregation.\n"
```
ludwiktrammer, line 96: "an API", "a SINGLE aggregation"

I would probably replace "query a database supporting" with "query a database with" but of course when it comes to the prompt, the results of the benchamarks is what's important.

src/dbally/iql_generator/prompt.py
```diff
@@ -85,3 +88,24 @@ def __init__(
...
+            "When prompted for just aggregation, use the following methods: \n"
```
ludwiktrammer, line 97: I don't understand what "When prompted for just aggregation" mean.

src/dbally/iql_generator/prompt.py
```diff
@@ -85,3 +88,24 @@ def __init__(
...
+            "{aggregations}"
+            "DO NOT INCLUDE arguments names in your response. Only the values.\n"
+            "You MUST use only these methods:\n"
+            "\n{aggregations}\n"
+            "It is VERY IMPORTANT not to use methods other than those listed above."
+            """If you DON'T KNOW HOW TO ANSWER DON'T SAY \"\" anything other than `UNSUPPORTED QUERY`"""
```
ludwiktrammer, line 103: The \"\" in the middle doesn't  make sense in the flow of this sentence

src/dbally/views/decorators.py
```diff
@@ -14,3 +14,18 @@ def wrapped(func: typing.Callable) -> typing.Callable:  # pylint: disable=missin
         return func
 
     return wrapped
+
+
+def view_aggregation() -> typing.Callable:
+    """
+    Decorator for marking a method as a filter
```
ludwiktrammer, line 21:
```suggestion
    Decorator for marking a method as an aggregation
```

src/dbally/views/pandas_base.py
```diff
@@ -36,6 +36,15 @@ async def apply_filters(self, filters: IQLQuery) -> None:
...
         self._filter_mask = await self.build_filter_node(filters.root)
 
+    async def apply_aggregation(self, aggregation: IQLQuery) -> None:
+        """
+        Applies the aggregation of choice to the view.
+
+        Args:
+            aggregation: IQLQuery object representing the aggregation to apply
+        """
+        pass  # pylint: disable=unnecessary-pass
```
ludwiktrammer, line 46: Pandas can support aggregations. Maybe at least mark it with TODO and add a ticket?

BTW, why you chose disabling the pylint rule instead of changing the code to be inline with it?

src/dbally/views/sqlalchemy_base.py
```diff
@@ -27,6 +28,15 @@ def get_select(self) -> sqlalchemy.Select:
         which will be used to build the query.
         """
 
+    def get_subquery(self) -> sqlalchemy.Subquery:
+        """
+        Creates the initial sqlalchemy.Subquery object, which will be used to build the query.
+
+        Returns:
+            The sqlalchemy.Subquery object based on private _select attribute.
```
ludwiktrammer, line 36: I don't think we should mention private elements (i.e., internal implementation details) in public documentation. On the other hand, I'm not sure the end-users need `get_subquery` at all, so maybe it should be marked as private as well (or removed and folded into the `apply_aggregation` method itself).

src/dbally/views/structured.py
```diff
@@ -94,6 +107,23 @@ async def apply_filters(self, filters: IQLQuery) -> None:
...
+    @abc.abstractmethod
+    def list_aggregations(self) -> List[ExposedFunction]:
+        """
+
+        Returns:
+            Aggregations defined inside the View.
+        """
+
+    @abc.abstractmethod
+    async def apply_aggregation(self, aggregation: IQLQuery) -> None:
```
ludwiktrammer, line 119: Aggregation can also be a single function call. We don't support IQL for aggregations. So this argument should be typed accordingly.

src/dbally/views/sqlalchemy_base.py
```diff
@@ -64,6 +74,16 @@ async def _build_filter_bool_op(self, bool_op: syntax.BoolOp) -> sqlalchemy.Colu
...
 
+    async def apply_aggregation(self, aggregation: IQLQuery) -> None:
+        """
+        Creates a subquery based on existing
+
+        Args:
+            aggregation: IQLQuery object representing the filters to apply
+        """
+        self._subquery = self.get_subquery()
+        self._subquery = await self.call_aggregation_method(aggregation.root)
```
ludwiktrammer, line 85: This seems very confusing. You assign a value to `self._subquery` only to overwrite it on the very next line be another value that is **not** a subquery object (just a normal `Select`).

I don't see why the class needs a `_subquery` property at all. I would remove `self._subquery` and replace the entire method with this line:

```
self._select = await self.call_aggregation_method(aggregation.root)
```

This assumes that the `call_aggregation_method` is responsible for creating the subquery obejct (which seems reasonable to me). Alternatively you can modify `call_aggregation_method` to take the subquery object as an argument.

This will also allow you to revert all the changes to the `execute` method.

src/dbally/views/sqlalchemy_base.py
```diff
@@ -77,13 +97,17 @@ def execute(self, dry_run: bool = False) -> ViewExecutionResult:
         """
 
         results = []
-        sql = str(self._select.compile(bind=self._sqlalchemy_engine, compile_kwargs={"literal_binds": True}))
+        statement = self._select
```
ludwiktrammer, line 100: As described in the comment to `apply_aggregation` - I don't think those changes are nessesary

ludwiktrammer (changes requested):

docs/quickstart/quickstart_code.py
```diff
@@ -55,6 +54,13 @@ def from_country(self, country: str) -> sqlalchemy.ColumnElement:
         """
         return Candidate.country == country
 
+    @decorators.view_aggregation()
+    def count_by_column(self, filtered_query: sqlalchemy.Select, column_name: str) -> sqlalchemy.Select:  # pylint: disable=W0602, C0116, W9011
```
micpst, line 58: For me it's too general, besides, as I understand we will print the whole signature to the LLM, how is it supposed to fill the `filter_query` argument? Also `column_name` will be a hallucination without proper context. I'd change it to something more specific and update the signature accordingly. `filtered_query` should be accessible via `self`.

src/dbally/prompt/aggregation.py
```diff
@@ -0,0 +1,122 @@
...
+        self,
+        question: str,
+        aggregations: List[ExposedFunction] = None,
+    ) -> None:
+        super().__init__()
+        self.question = question
+        self.aggregations = "\n".join([str(aggregation) for aggregation in aggregations]) if aggregations else []
+
+
+class AggregationFormatter:
```
micpst, line 46: This class is generator to me. I'd put it in `iql_generator` module.

PatrykWyzgowski: These changes were suggested by @ludwiktrammer and as the aggregations are not IQL-generated directly I decided to keep it in a separate, dedicated module.

docs/quickstart/quickstart_code.py
PatrykWyzgowski: `filtered_query` is a hidden argument. Please see: https://github.com/deepsense-ai/db-ally/blob/pw/add-single-aggregation/src/dbally/views/methods_base.py#L18

docs/quickstart/quickstart_code.py
```diff
@@ -55,6 +54,12 @@ def from_country(self, country: str) -> sqlalchemy.ColumnElement:
         """
         return Candidate.country == country
 
+    @decorators.view_aggregation()
+    def count_by_column(self, subquery: sqlalchemy.Select, column_name: str) -> sqlalchemy.Select:  # pylint: disable=W0602, C0116, W9011
```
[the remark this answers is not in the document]
PatrykWyzgowski: First comment - resolved. The latter to be handled in separate issue.

docs/quickstart/quickstart_code.py
micpst: ok, but why its here, if it can be accessed via `self`?

src/dbally/prompt/aggregation.py
micpst: Aggregations are part of the IQL syntax, and the fact that we generate them separately doesn't exclude them from the IQL language. This class reflects the behavior of `IQLGenerator`, so copying this logic to the `prompt` module is not correct for me. Perhaps we should have 2 IQL generators - `FilterGenerator` and `AggregationGenerator` to make it cleaner.

docs/quickstart/quickstart_code.py
PatrykWyzgowski: https://github.com/deepsense-ai/db-ally/pull/62/files/2918ba5a81d0da7e3247964b5d792f3a7c1ee080#diff-c9a06e39afec54ec8db60598f66db7326f0ff226303b29e8d7e211960a77761bR130
See how this parameter is injected.

src/dbally/prompt/aggregation.py
PatrykWyzgowski: @mhordynski - what is your perspective from architectural point of view?

mhordynski: I agree with having `IQLGenerator` and separating later to `IQLFiltersGenerator` / `IQLAggegationGenerator`. from my perspective aggregations, along other features like `Context()` are becoming parts of the IQL

src/dbally/prompt/aggregation.py
```diff
@@ -0,0 +1,122 @@
...
+        agg = formatted_prompt.response_parser(response)
+        # TODO: Move IQL query parsing to prompt response parser
+        return await IQLQuery.parse(
+            source=agg,
+            allowed_functions=aggregations or [],
+            event_tracker=event_tracker,
+        )
+
+
+AGGREGATION_GENERATION_TEMPLATE = PromptTemplate[AggregationPromptFormat](
```
micpst, line 103: I think we should first ask the LLM whether to use aggregation at all before asking it to generate it. In this implementation, there is no way to skip this step, which introduces hallucinated answers or `UnsupportedAggregationError` for filter-only questions.

src/dbally/iql_generator/iql_generator.py
```diff
@@ -15,57 +20,151 @@
...
-    It uses LLM to generate text-based responses, passing in the prompt template, formatted filters, and user question.
+        Returns:
+            True if the generation failed, False otherwise.
+        """
+        return isinstance(self.filters, Exception) or isinstance(self.aggregation, Exception)
+
+
+class IQLGenerator:
+    """
+    Program that orchestrates all IQL operations for the given question.
```
ludwiktrammer, line 48: nit(non-blocking): this doesn't conform to my conception of what "program" is - for me "program" is a standalone piece of software that can be run by itself. I would just call this "class" or would loose the noun completely.

src/dbally/iql_generator/prompt.py
```diff
@@ -71,82 +112,131 @@ def __init__(self, *, question: str, examples: List[FewShotExample] = None) -> N
...
-                """If you DON'T KNOW HOW TO ANSWER DON'T SAY \"\", SAY: `UNSUPPORTED QUERY` INSTEAD! """
-                "This is CRUCIAL, otherwise the system will crash. "
+                "Given a question, determine whether the answer requires initial data filtering in order to compute it.\n"
+                "Initial data filtering is a process in which the result set is reduced to only include the rows "
+                "that meet certain criteria specified in the question.\n\n"
+                "---\n\n"
+                "Follow the following format.\n\n"
+                "Question: ${{question}}\n"
+                "Hint: ${{hint}}"
+                "Reasoning: Let's think step by step in order to ${{produce the decision}}. We...\n"
```
ludwiktrammer, line 151: I don't get the "We..." part. I also don't understand `${{produce the decision}}`. Is it replaced with something? This doesn't look like a proper syntax for that

src/dbally/iql_generator/prompt.py
```diff
@@ -71,82 +112,131 @@ def __init__(self, *, question: str, examples: List[FewShotExample] = None) -> N
...
+        methods: List[ExposedFunction],
+        examples: Optional[List[FewShotExample]] = None,
     ) -> None:
         """
         Constructs a new IQLGenerationPromptFormat instance.
 
         Args:
             question: Question to be asked.
-            filters: List of filters exposed by the view.
+            methods: List of filters exposed by the view.
```
ludwiktrammer, line 130: I assume you changed "filters" to "methods" because this can also take aggregation methods (?) but the description stills says "filters"

src/dbally/iql_generator/prompt.py
```diff
@@ -71,82 +112,131 @@ def __init__(self, *, question: str, examples: List[FewShotExample] = None) -> N
...
             examples: List of examples to be injected into the conversation.
+            aggregations: List of aggregations exposed by the view.
         """
         super().__init__(examples)
         self.question = question
-        self.filters = "\n".join([str(filter) for filter in filters])
+        self.methods = "\n".join([str(condition) for condition in methods]) if methods else []
```
ludwiktrammer, line 136: So the type of `self.methods` can either by a string or (an empty) list? Doesn't sound right - it should be either one or the other.

As a side note, you don't need to create the temporary list inside join - `"\n".join(str(condition) for condition in methods)` would work just fine.

src/dbally/views/methods_base.py
```diff
@@ -99,3 +109,19 @@ async def call_filter_method(self, func: syntax.FunctionCall) -> Any:
         if inspect.iscoroutinefunction(method):
             return await method(*args)
         return method(*args)
+
+    async def call_aggregation_method(self, func: syntax.FunctionCall) -> Any:
```
ludwiktrammer, line 113: Do we need a separate `call_aggregation_method`? It seems that we could just rename `call_filter_method` to `call_method` since it seems that both the body, the documentation, and the purpose of both is the same.

src/dbally/views/pandas_base.py
```diff
@@ -19,48 +21,59 @@ class DataFrameBaseView(MethodsBaseView):
...
 
-    async def build_filter_node(self, node: syntax.Node) -> pd.Series:
+    async def apply_aggregation(self, aggregation: IQLAggregationQuery) -> None:
+        """
+        Applies the aggregation of choice to the view.
+
+        Args:
+            aggregation: IQLQuery object representing the aggregation to apply.
+        """
+        self._groupbys, self._aggregations = await self.call_aggregation_method(aggregation.root)
```
ludwiktrammer, line 51: Looking at how those aggregation methods look like, for example:

```
    @view_aggregation()
    def mean_age_by_city(self) -> Tuple[str, List[Tuple[str, str]]]:
        return "city", [("age", "mean")]
```

It's very hard to tell looking at such code what "city" is, what "age" is, and what "mean" is. Maybe the return type should some sort of a special aggregation data class / named tuple? So that the individual elements are defined with names, showing what they are.

src/dbally/iql_generator/prompt.py
```diff
@@ -71,82 +112,131 @@ def __init__(self, *, question: str, examples: List[FewShotExample] = None) -> N
...
+                "(Respond with True or False)\n\n"
             ),
         },
         {
             "role": "user",
-            "content": "{question}",
+            "content": (
+                "Question: {question}\n"
+                "Hint: Look for words indicating data specific features.\n"
+                "Reasoning: Let's think step by step in order to "
```
ludwiktrammer, line 161: It's not clear to me why `Hint` and `Reasoning` are repeated in System and User message, although in System they look dynamic (thought I'm not sure where the values come from) and in User they look static.

ludwiktrammer (commented):
Looks very good in general - I have just a few small questions

src/dbally/iql_generator/prompt.py
micpst: This is how `dspy` implements the `ChainOfThought` prompt. This `${{something}}` syntax is used to define dynamic content for LLM.

src/dbally/iql_generator/prompt.py
micpst: We use system prompt to define a task and a structure of the conversation. Some parts of the prompts are static, some are dynamic. Here we tell the LLM to follow the format and we help it maintain the structure in the user prompt, so the LLM knows how to generate reasoning based on the hint we have already provided. There is only one user input attribute and it is `{question}`, the double brackets `{{}}` indicate that this is an attribute for the LLM to fill in, the hint is already provided, so the LLM will start generating with a fixed hint and therefore should achieve better accuracy. 

I've tested this on our benchmark and prompts with hints on CoT performs better in general on smaller models, since they require more structure.

src/dbally/views/pandas_base.py
micpst: Good idea, will implement this.

src/dbally/iql_generator/prompt.py
micpst: Thanks for catching, will fix this

src/dbally/views/methods_base.py
micpst: I like the separate method here because we are also doing the decorator validation here, which is specific to the aggregation method. We could add that as an argument here, but it would make the interface a bit ambiguous, imho. But maybe I could create a private method inside the `MethodsBasedView` just for method calls, since this is a common part and use it in `call_filter_method` and `call_aggregation_method`.

src/dbally/iql_generator/iql_generator.py
micpst: That's fair, I can change it

src/dbally/iql_generator/prompt.py
ludwiktrammer: so the `${{placeholders}` are for the LLM to interpret? Looks weird to me but if it works - it works 🤷

src/dbally/views/structured.py
```diff
@@ -19,21 +17,18 @@
 
 class BaseStructuredView(BaseView):
     """
-    Base class for all structured [Views](../../concepts/views.md). All classes implementing this interface has\
+    Base class for all structured views. All classes implementing this interface has\
```
ludwiktrammer, line 20: nit: why remove the link?

src/dbally/views/sqlalchemy_base.py
```diff
@@ -15,28 +15,42 @@ class SqlAlchemyBaseView(MethodsBaseView):
...
 
     @abc.abstractmethod
     def get_select(self) -> sqlalchemy.Select:
-        r"""
-        Creates the initial
-        [SqlAlchemy select object
-        ](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select)
-        which will be used to build the query.
         """
+        Creates initial SELECT statement for the view.
```
ludwiktrammer, line 31: nit: why remove the link? Since we ask people to create sql alchemy statements I believe it's important that we give them link to resources about creating them

ludwiktrammer (approved):

src/dbally/views/structured.py
micpst: this link didn't work

src/dbally/views/sqlalchemy_base.py
micpst: I agree, but this link was badly formatted because of the line limitation and did not render correctly. I decided to remove it because if the user wants to use the SQL alchemy view, he should know the basic sqlalchemy API.

ludwiktrammer: I definitely don't agree with "if the user wants to use the SQL alchemy view" - this is currently the only way to use db-ally with an SQL database that we support, many users coming to db-ally will have to learn both at the same time, and we should provide them the resources to do so.

src/dbally/views/structured.py
ludwiktrammer: Just tested the link again (using the `mkdocs serve` command and looking at the description of the class under http://127.0.0.1:8000/reference/views/structured/) and the link works for me

src/dbally/views/sqlalchemy_base.py
ludwiktrammer: It seems to render correctly in my tests. What should I do to reproduce the problem?

src/dbally/views/structured.py
micpst: In the documents it works fine, but in the IDE it doesn't work. I wonder if we really need this if we are linking to the same resource referenced in the tips section.

src/dbally/views/sqlalchemy_base.py
micpst: I rechecked and now I don't have problem with this, so I think I can reverse the change here

src/dbally/views/structured.py
ludwiktrammer: We have similar links to specific concepts all over the documentations - we specifically worked on adding them when originally writing the documentation, so that the users can quickly reference what a given concept is about. If you think we should remove them - we can discuss this. But I don't think quietly removing just a single instance makes sense.

PR #64, opened 2024-06-24
Feat: Implement global event handlers

karllu3:


examples/visualize_views_code.py
```diff
@@ -5,15 +5,17 @@
 from recruiting.cypher_text2sql_view import SampleText2SQLViewCyphers, create_freeform_memory_engine
 
 import dbally
-from dbally.audit import CLIEventHandler
+from dbally.audit.event_handlers.buffer_event_handler import BufferEventHandler
```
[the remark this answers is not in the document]
patryk-wyzgowski: Let's consider `litellm` as an inspiration.

src/dbally/global_handlers.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Raises:
+        ValueError: If the handler is not an instance of EventHandler.
+    """
+    if isinstance(type(event_handler), EventHandler):
+        raise ValueError(f"{event_handler} is not an instance of EventHandler")
+    global_event_handlers_list.append(event_handler)
+
+
+def find_event_handler(object_type: Type) -> Optional[EventHandler]:
```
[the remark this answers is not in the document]
karllu3: I would avoid singleton for handling global variables.

src/dbally/__init__.py
```diff
@@ -21,29 +23,32 @@
 from .exceptions import DbAllyError
 from .llms.clients.exceptions import LLMConnectionError, LLMError, LLMResponseError, LLMStatusError
 
+global_event_handlers: List[Callable] = []
```
mhordynski, line 26: Let's name it just `event_handlers`

src/dbally/gradio/gradio_interface.py
```diff
@@ -28,6 +28,23 @@ async def create_gradio_interface(user_collection: Collection, preview_limit: in
...
+
+    This function iterates over the list of global event handlers stored in `dbally.global_event_handlers`.
+    It checks the type of each handler, and if it finds one that is an instance of `BufferEventHandler`, it
+    returns that handler. If no such handler is found, the function returns `None`.
+
+    Returns:
+        The first instance of `BufferEventHandler` found in the list, or `None` if no such handler is found.
+    """
+    for handler in dbally.global_event_handlers:
+        if type(handler) is BufferEventHandler:  # pylint: disable=C0123
```
mhordynski, line 43: Why not `isinstance(...)`?

karllu3: Due to inheritence 
```isinstance(handler, BufferEventHandler)``` would also return true for CLIEventHandler.
BufferEventHandler inherits from CLIEventHandler

File: src/dbally/__init__.py (hunk 3 above)
Comment (karllu3) [reply src/dbally/__init__.py:26]: I called it this way to not overwrite it by accident by local variable.
Regarding overwriting, collection module event_handler is also name of the ask parameter. 
When we use global with prefix dbally.event_handler overwrite would not happens but I thought it could be tricky if someone won't refer to it like that.

File: src/dbally/gradio/gradio_interface.py (hunk 4 above)
Comment (mhordynski) [reply src/dbally/gradio/gradio_interface.py:43]: Are you sure? ;) As `BufferEventHandler` inherits from CLI handler (and not the other way around) isinstance won't return `True` in that case

File: src/dbally/__init__.py (hunk 3 above)
Comment (mhordynski) [reply src/dbally/__init__.py:26]: Maybe we should rename the local variable to something like `override_event_handlers`? That way, it's clear that we're providing handlers meant to override the existing ones.

File: src/dbally/gradio/gradio_interface.py [hunk 5]
```diff
@@ -127,10 +162,12 @@ async def _ui_ask_query(
             generated_query = {"Query": "No view matched to query"}
             data = pd.DataFrame()
         finally:
+            print(f" ask log 1 {self.log}")
```
micpst: Please get rid of this debug statement here.

src/dbally/gradio/gradio_interface.py
```diff
@@ -127,10 +162,12 @@ async def _ui_ask_query(
...
             self.log.seek(0)
             log_content = self.log.read()
 
-        gradio_dataframe, empty_dataframe_warning = self._load_gradio_data(data, "Results", "No matching results found")
+        gradio_dataframe, empty_dataframe_warning = self._load_gradio_data(data, "Results")
+        print(f" ask log 2 {self.log}")
```
micpst: And here

src/dbally/collection/collection.py
```diff
@@ -58,17 +59,24 @@ def __init__(
...
         self._view_selector = view_selector
         self._nl_responder = nl_responder
-        self._event_handlers = event_handlers
         self._llm = llm
 
+        if not event_handlers:
+            event_handlers = dbally.global_event_handlers
+        elif event_handlers != dbally.global_event_handlers:
+            # At this moment, there is no event tracker initialized to record an event
+            print(f"WARNING: Default event handler has been overwritten for {self.name}.")
```
micpst, line 69: Lets use `warnings.warn`

src/dbally/_main.py
```diff
@@ -36,7 +38,7 @@ def create_collection(
         name: Name of the collection is available for [Event handlers](event_handlers/index.md) and is\
         used to distinguish different db-ally runs.
         llm: LLM used by the collection to generate responses for natural language queries.
-        event_handlers: Event handlers used by the collection during query executions. Can be used to\
+        override_event_handlers: Event handlers used by the collection during query executions. Can be used to\
```
micpst, line 41: Let's call it `event_handlers` and mention here that it will override global handlers defined elsewehere.

src/dbally/collection/collection.py
```diff
@@ -276,3 +276,60 @@ async def update_similarity_indexes(self) -> None:
...
+
+    Returns:
+        a new instance of db-ally Collection
+
+    Raises:
+        ValueError: if default LLM client is not configured
+    """
+    view_selector = view_selector or LLMViewSelector(llm=llm)
+    nl_responder = nl_responder or NLResponder(llm=llm)
+    event_handlers = event_handlers or dbally.event_handlers
```
micpst, line 327: This is not needed here, because we set event handlers in the `Collection` initializer.

src/dbally/__init__.py
```diff
@@ -21,29 +22,32 @@
 from .exceptions import DbAllyError
 from .llms.clients.exceptions import LLMConnectionError, LLMError, LLMResponseError, LLMStatusError
 
+event_handlers: List[Callable] = []
```
micpst, line 25: Wrong typing

micpst (approved):

Issue #54, opened 2024-06-18
feat: request contextualization


mhordynski:
### Feature description

db-ally should allow the provision of extra context for its runs, such as identifying the user who is asking the question. This context can be particularly useful in StructuredViews to craft responses in a personalized manner, taking into consideration parameters like the identity of the person asking the question or other relevant details.

### Motivation

Given the following filters for searching employee database:

```
filter_by_department(department: str)
filter_by_skill(skill: str)
```

and user question like this:

```
Find me employees from my department that are skilled in Python
```

db-ally should be able to fecth user department from injected context and answer the question.

### Additional context

We probably need to design how we want to indicate in IQL that we wish to use context rather than an explicit value. For example:

```python
@decorators.view_filter()
def filter_by_department(self, department: str | SelfKeyword):
    ...
```

and later in IQL LLM could call filter like this:

```filter_by_department(SELF)```

PR #65, opened 2024-06-26
Request contextualisation - core functionality

jcierocki:
Solves #54

### DIFF
--- a/src/dbally/context/context.py
+++ b/src/dbally/context/context.py
@@ -1,0 +1,2 @@
+class BaseCallerContext:
+    pass
--- a/src/dbally/context/utils.py
+++ b/src/dbally/context/utils.py
@@ -26,6 +26,9 @@ def _extract_params_and_context(
     context = None
     # TODO confirm whether to use typing.get_type_hints(method) or method.__annotations__
     for name_, type_ in typing.get_type_hints(filter_method_).items():
+        if name_ in hidden_args:
+            continue
+
         if isclass(type_) and issubclass(type_, BaseCallerContext):
             # this is the case when user provides a context but no other type hint for a specifc arg
             # TODO confirm whether this case should be supported
--- a/src/dbally/context/utils.py
+++ b/src/dbally/context/utils.py
@@ -1,29 +1,51 @@
 import typing
 
-from .context import BaseCallerContext
+from inspect import isclass
+
+from dbally.context.context import BaseCallerContext
+from dbally.views.exposed_functions import MethodParamWithTyping
 
 
-def _extract_filter_context(filter_: typing.Callable) -> typing.Optional[typing.Type[BaseCallerContext]]:
+def _extract_params_and_context(
+    filter_method_: typing.Callable, hidden_args: typing.List[str]
+) -> typing.Tuple[
+    typing.List[MethodParamWithTyping],
+    typing.Optional[typing.Type[BaseCallerContext]]
+]:
     """
-    Extracts a SINGLE caller's context given a StructuredView filter.
+    Processes the MethodsBaseView filter method signauture to extract the args and type hints in the desired format. Context claases are getting excluded the returned MethodParamWithTyping list. Only the first BaseCallerContext class is returned.
 
     Args:
-        filter_: MethodsBaseView filter method (annotated with @decorators.view_filter() decorator)
+        filter_method_: MethodsBaseView filter method (annotated with @decorators.view_filter() decorator)
 
     Returns:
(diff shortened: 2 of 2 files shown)


src/dbally/context/context.py
```diff
@@ -0,0 +1,59 @@
...
+from pydantic import BaseModel
+from typing_extensions import Self, TypeAlias
+
+from dbally.context.exceptions import ContextNotAvailableError
+
+# CustomContext = TypeVar('CustomContext', bound='BaseCallerContext', covariant=True)
+CustomContext: TypeAlias = "BaseCallerContext"
+
+
+class BaseCallerContext(BaseModel):
```
mhordynski, line 13: pydantic is not a required dependency at the moment, and we should avoid it if possible (v1 vs v2 issues). Maybe a bare dataclass is enough here?

mhordynski: We can still support pydantic models as context, but we shouldn't enforce it

src/dbally/iql_generator/prompt.py
```diff
@@ -74,6 +74,10 @@ def __init__(
                 "You MUST use only these methods:\n"
                 "\n{filters}\n"
                 "It is VERY IMPORTANT not to use methods other than those listed above."
+                "Finally, if a called function argument value is not directly specified in the query but instead requires knowledge of some additional context, than substitute that argument value by: BaseCallerContext()."
```
mhordynski, line 77: maybe just `Context()`?

jcierocki: Right. I will adjust it.

src/dbally/context/context.py
jcierocki: Fixed, thanks :)

src/dbally/iql_generator/prompt.py
jcierocki: Done. Thanks :)

src/dbally/context/_utils.py
```diff
@@ -0,0 +1,75 @@
...
+from dbally.views.exposed_functions import MethodParamWithTyping
+
+ContextClass: type_ext.TypeAlias = Optional[Type[BaseCallerContext]]
+
+
+def _extract_params_and_context(
+    filter_method_: type_ext.Callable, hidden_args: Sequence[str]
+) -> Tuple[Sequence[MethodParamWithTyping], ContextClass]:
+    """
+    Processes the MethodsBaseView filter method signauture to extract the args and type hints in the desired format.
```
ludwiktrammer, line 16: Typos: "signauture", "claases", "are getting excluded the returned"

src/dbally/context/context.py
```diff
@@ -0,0 +1,64 @@
...
+
+class BaseCallerContext(ABC):
+    """
+    An interface for contexts that are used to pass additional knowledge about
+    the caller environment to the filters. LLM will always return `Context()`
+    when the context is required and this call will be later substituted by an instance of
+    a class implementing this interface, selected based on the filter method signature (type hints).
+    """
+
+    _alias: str = "Context"
```
ludwiktrammer, line 20: nit: instead of providing a default value, I would require inheriting classes to define their own (throw exception otherwise or even better use an abstract method definition), to ensure that they don't use the default value by accident. Especially since it might be easy for people to assume that the name of the class itself is used as an alias automatically.

src/dbally/context/_utils.py
jcierocki: Thanks.

src/dbally/iql_generator/iql_generator.py
```diff
@@ -53,14 +55,16 @@ async def generate_iql(
...
+                an instance of a subclass of BaseCallerContext.
 
         Returns:
             Generated IQL query.
         """
         prompt_format = IQLGenerationPromptFormat(
             question=question,
             filters=filters,
-            examples=examples,
+            examples=examples or [],
```
micpst, line 67: This is handled in the `PromptFormat` initializator.

src/dbally/context/exceptions.py
micpst, line 1: All dbally exceptions should inherit from `dbally.exceptions.DbAllyError`. To be consistent with the rest of the code base:
- `BaseContextException` should inherit from `DbAllyError`
- rename `BaseContextException` -> `ContextError`
- use inheritance for `ContextNotAvailableError` and `ContextualisationNotAllowed`

micpst (commented):
Please update the docs on how contexts are defined and how they work.

PR #70, opened 2024-07-05
feat: Add OpenTelemetry event handler

BartMiki:


src/dbally/audit/event_handlers/otel_event_handler.py
```diff
@@ -0,0 +1,134 @@
...
+    async def request_end(self, output: RequestEnd, request_context: Span) -> None:
+        """
+        Finalizes entire request, ending the span for this request.
+
+        Args:
+            output: output generated for this request
+            request_context: span to be closed
+        """
+        if self.record_outputs:
+            request_context.set_attribute("db-ally.response", output.result.textual_response)
```
mhordynski, line 127: textual_response may not be present (nl_responder is optional). 
Maybe we should add context here (generated IQL, SQL or other query)

BartMiki: Sure, I will add this

mhordynski (approved):
looks good :raised_hands:

PR #78, opened 2024-08-05
feat(views): optional filtering for structured views

micpst:
This PR adds a prompt for the LLM to decide whether to apply filtering to the question.


mhordynski (approved):
lgtm

